PR #89, opened 2024-03-15
RHIDP-1247: Dynamic plugin Helm install and RHIDP-1487: Chapter 5 restructure

linfraze:
Associated Jira tickets: 
RHIDP-1247
RHIDP-1487

rh-pre-commit.version: 2.2.0
rh-pre-commit.check-secrets: ENABLED


hmanwani-rh:
@linfraze Please resolve conflicts in the PR.

modules/rhdh-plugins-reference/ref-rhdh-plugins.adoc
```diff
@@ -10,9 +10,9 @@ For more information on Red Hat Technology Preview features, see https://access.
 Additional detail on how Red Hat provides support for bundled community dynamic plugins is available on the https://access.redhat.com/policy/developerhub-support-policy[Red Hat Developer Support Policy] page.
 ====
 
-There are 47 plugins.
+There are more than 45 plugins available in {product}. See the xref:rhdh-supported-plugins[Dynamic plugins support matrix] for details.
```
Gerry-Forde, line 13: I'd suggest stating the actual number of available plugins rather than saying "more than 45", I know it takes a bit more maintenance but maybe some users want to know exactly how many. This will change again before GA as Nick has a PR adding 9 more plugins, so I think it will be 56.

modules/rhdh-plugins-reference/ref-rhdh-plugins.adoc
```diff
@@ -1,5 +1,5 @@
 [id="rhdh-supported-plugins"]
-== Dynamic plugins included in {product}
+= Preinstalled dynamic plugin descriptions and details
```
Gerry-Forde, line 2: Might "Preinstalled" give the impression that the plugins are installed and enabled? I know you discussed this with SME, so I'm happy to go with whatever terminology you decided was most appropriate.

modules/rhdh-plugins-reference/proc-install-external-dynamic-plugins-helm.adoc
```diff
@@ -0,0 +1,16 @@
...
+
+To install dynamic plugins in {product-short}, add the following `global.dynamic` parameters in your Helm Chart:  
+
+* `plugins`: the dynamic plugins list intended for installation. By default, the list is empty. You can populate the plugins list with the following fields:
+** `package`: a package specification for the dynamic plugin package that you want to install. You can use a package for either an internal or external dynamic plugin installation. For an internal installation, use from a local path to the folder containing the dynamic plugin for a local installation. For an external installation, use a or from package specification from a public NPM repository.
+** `integrity` (required for external packages): an integrity checksum in the form of `<alg>-<digest>` specific to the package. Supported algorithms include `sha256`, `sha384` and `sha512`. 
+** `pluginConfig`: an optional plugin-specific app-config yaml fragment. See plugin configuration for more information.details.
+** `disabled``: disables the dynamic plugin if set to `true``. Default: `false`.
```
Gerry-Forde, line 13:
```suggestion
** `disabled`: disables the dynamic plugin if set to `true`. Default: `false`.
```

modules/rhdh-plugins-reference/proc-install-external-dynamic-plugins-helm.adoc
```diff
@@ -0,0 +1,16 @@
...
+* `includes`: a list of YAML files utilizing the same syntax. 
+
+[NOTE] The `plugins` list in the `includes` file is merged with the `plugins` list in the main Helm values. If a plugin package is mentioned in both `plugins` lists, the `plugins` fields in the main Helm values override the `plugins` fields in the `includes` file. The default configuration includes the `dynamic-plugins.default.yaml` file, which contains all of the dynamic plugins preinstalled in {product-short}, whether enabled or disabled by default.
```
Gerry-Forde, line 16:
```suggestion
+
[NOTE] 
The `plugins` list in the `includes` file is merged with the `plugins` list in the main Helm values. If a plugin package is mentioned in both `plugins` lists, the `plugins` fields in the main Helm values override the `plugins` fields in the `includes` file. The default configuration includes the `dynamic-plugins.default.yaml` file, which contains all of the dynamic plugins preinstalled in {product-short}, whether enabled or disabled by default.
```

modules/rhdh-plugins-reference/ref-rhdh-plugins.adoc
linfraze: Yes, I was definitely trying to keep us from being wrong/out of date the moment they add another plugin, but I can change this back to the exact number.

modules/rhdh-plugins-reference/ref-rhdh-plugins.adoc
linfraze: Good Q! I made this change because the IBM style guide says to use "preinstalled" instead of "pre-loaded" and I wanted us to be compliant, but I do see what you mean. I did talk to both SMEs about these terms because the difference between "included" and "preinstalled" in not clear. Rather than rely on the nuances between "preinstalled" and "included" to communicate preinstalled and enabled/disabled by default, we are starting to more clearly and directly say what these words mean, which is also helpful for translation.

linfraze: In multiple places, we say that plugins are enabled or disabled by default, but we should prob include this info even more directly in `ref-rhdh-plugins` and 'con-preinstalled-dynamic-plugins`

modules/rhdh-plugins-reference/ref-rhdh-plugins.adoc
Gerry-Forde: Plugins will only be added in each quarterly release, so it's probably a maintenance overhead that we can live with. Also in future releases, I think the plan might be to automate the generation of this table content, so we may not have to worry about it.

hmanwani-rh:
@linfraze Please rebase your branch

PR #115, opened 2024-03-20
RHIDP-2147 chore: Added editorconfig for consistent IDE behavior

themr0c:
This update specifies editor configuration for consistent IDE behavior.

See: https://editorconfig.org/

https://github.com/redhat-developer/red-hat-developers-documentation-rhdh/assets/243761/10d8fd54-adf1-474f-8321-10ceec29bdff


nickboldt (approved):
seems OK, if we all like tabs to be 2 spaces wide. Assuming that's ok from an accessibility perspective or that individual people can override that

PR #137, opened 2024-04-02
Fixed the service now package to match the correct name

jayfray12:
@hmanwani-rh here is the cherry pick for 1.0


hmanwani-rh:
@nickboldt Could you assist here? I'm unsure why the build check isn't functioning properly.

PR #162, opened 2024-04-11
Add fix for deploy action failure

abrennan89:
One of my manual cherrypicked PRs has a check failing for deploying the preview: https://github.com/redhat-developer/red-hat-developers-documentation-rhdh/actions/runs/8648296395/job/23711771598?pr=161

image

It seems to be happening occassionally for other PRs also, so I tried to look into a fix for it. Based on https://github.com/peaceiris/actions-gh-pages/issues/339 I am proposing that updating the variable here from `github_token` to `deploy_key` may fix the issue, but I am not an expert on GH actions, so will need someone to sanity check 🙂

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -23,7 +23,7 @@ on:
   # will prevent the job from running until it's approved manually by human intervention.
   pull_request_target:
     types: [opened, synchronize, reopened, ready_for_review]
-    branches: 
+    branches:
     - main
     - rhdh-1.**
     - 1.**.x
@@ -67,7 +67,7 @@ jobs:
         run: |
           # update
           sudo apt-get update -y || true
-          # install 
+          # install
           sudo apt-get -y -q install asciidoctor && asciidoctor --version
           echo "GIT_BRANCH=${GITHUB_HEAD_REF:-${GITHUB_REF#refs/heads/}}" >> $GITHUB_ENV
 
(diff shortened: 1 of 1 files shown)


.github/workflows/pr.yml
```diff
@@ -81,19 +81,19 @@ jobs:
...
+          deploy_key: ${{ secrets.RHDH_BOT_TOKEN }}
           publish_branch: gh-pages
           keep_files: true
           publish_dir: ./titles-generated
 
       - name: PR comment with doc preview
         uses: actions/github-script@v7
         with:
-          github-token: ${{ secrets.RHDH_BOT_TOKEN }}
+          deploy_key: ${{ secrets.RHDH_BOT_TOKEN }}
```
nickboldt, line 92:
```suggestion
          github-token: ${{ secrets.RHDH_BOT_TOKEN }}
```

See https://github.com/actions/github-script?tab=readme-ov-file#using-a-separate-github-token -- deploy_token is not supported for this action.

.github/workflows/pr.yml
```diff
@@ -81,7 +81,7 @@ jobs:
         uses: peaceiris/actions-gh-pages@v3
         # if: github.ref == 'refs/heads/main'
         with:
-          github_token: ${{ secrets.RHDH_BOT_TOKEN }}
+          deploy_key: ${{ secrets.RHDH_BOT_TOKEN }}
```
nickboldt, line 84: we can try this if you want -- but I'm not sure if we're using https or ssh to check out sources.

https://github.com/peaceiris/actions-gh-pages?tab=readme-ov-file#supported-tokens

abrennan89: Ah, I see. Is there a way to check?

.github/workflows/pr.yml
```diff
@@ -81,7 +81,7 @@ jobs:
         uses: peaceiris/actions-gh-pages@v3
```
nickboldt, line 81:
```suggestion
        uses: peaceiris/actions-gh-pages@v4
```

.github/workflows/pr.yml
nickboldt: try it and see if it breaks; then revert if so :D

PR #205, opened 2024-04-29
RHIDP-2208: Create assemblies directory

linfraze:
Creates a directory that will later contain all of the repo's assembly files. 
It is standard practice for the assemblies dir to also contain a symlink to the modules dir.


Gerry-Forde (approved):
I haven't encountered this practice before, I suppose it's just for convenience?

linfraze:
@Gerry-Forde The symlink to the modules folder is needed because assembly files `include` module files

PR #201, opened 2024-04-26
RHIDP-1541: Adding docs for creating templates

abrennan89:
**Versions:**
1.1+

**Jira link:**
- https://issues.redhat.com/browse/RHIDP-1541

**Preview link:**
https://redhat-developer.github.io/red-hat-developers-documentation-rhdh/pr-201/admin-rhdh/#assembly-admin-templates

**Reviews:**
- [x] SME: @cmoulliard @jayfray12 
- [x] QE: @zdrapela @albarbaro  
- [x] Peer review: @linfraze 

**Additional information:**
There are separate Jiras and will be separate PRs for adding templates and custom filters. See:
* https://issues.redhat.com/browse/RHIDP-1542
* https://issues.redhat.com/browse/RHIDP-2248
These are outside the scope of this PR.

PR #219, opened 2024-05-07
RHIDP-1542: Document adding templates

abrennan89:
**Versions:**
1.1+

**Jira link:**
- https://issues.redhat.com/browse/RHIDP-1542

**Preview link:**
https://redhat-developer.github.io/red-hat-developers-documentation-rhdh/pr-219/admin-rhdh/#proc-adding-templates_assembly-admin-templates

**Reviews:**
- [x] SME: @jayfray12 @cmoulliard 
- [x] QE: @Omar-AlJaljuli 
- [x] Peer review: @linfraze

### DIFF
--- a/modules/templates/proc-adding-templates.adoc
+++ b/modules/templates/proc-adding-templates.adoc
@@ -1,0 +1,39 @@
+// Module included in the following assemblies:
+//
+// * assemblies/assembly-admin-templates.adoc
+
+:_mod-docs-content-type: PROCEDURE
+[id="proc-adding-templates_{context}"]
+= Importing an existing template to {product}
+
+You can add an existing template to your {product} instance by using the Catalog Processor.
+
+.Prerequisites
+
+* You have created a directory or repository that contains at least one template YAML file.
+* If you want to use a template that is stored in a repository such as GitHub or GitLab, you must configure a {product} integration for your provider.
+
+.Procedure
+
+* In the `app-config.yaml` configuration file, modify the `catalog.rules` section to include a rule for templates, and configure the `catalog.locations` section to point to the template that you want to add, as shown in the following example:
++
+[source,yaml]
+----
+# ...
+catalog:
+  rules:
+    - allow: [Template] # <1>
+  locations:
+    - type: url # <2>
+      target: https://<repository_url>/example-template.yaml # <3>
+# ...
+----
+<1> To allow new templates to be added to the catalog, you must add a `Template` rule.
+<2> If you are importing templates from a repository, such as GitHub or GitLab, use the `url` type.
+<3> Specify the URL for the template.
+
+.Verification
+
+. Click the *Catalog* tab in the navigation panel.
+. In the *Kind* drop-down menu, select *Template*.
+. Confirm that your template is shown in the list of existing templates.
--- a/modules/templates/proc-creating-templates.adoc
+++ b/modules/templates/proc-creating-templates.adoc
@@ -1,14 +1,12 @@
 // Module included in the following assemblies:
 //
-// * admin/assembly-admin-templates.adoc
+// * assemblies/assembly-admin-templates.adoc
 
 :_mod-docs-content-type: PROCEDURE
 [id="proc-creating-templates_{context}"]
 = Creating a template by using the Template Editor
 
 You can create a template by using the Template Editor.
-
-
 
 .Procedure
 
--- a/assemblies/assembly-admin-templates.adoc
+++ b/assemblies/assembly-admin-templates.adoc
@@ -17,3 +17,10 @@ include::modules/templates/ref-creating-templates.adoc[leveloffset=+1]
 * link:https://backstage.io/docs/features/software-templates/writing-templates[Backstage documentation - Writing Templates]
 * link:https://backstage.io/docs/features/software-templates/builtin-actions[Backstage documentation - Builtin actions]
 * link:https://backstage.io/docs/features/software-templates/writing-custom-actions[Backstage documentation - Writing Custom Actions]
+
+include::modules/templates/proc-adding-templates.adoc[leveloffset=+1]
+
+[role="_additional-resources"]
+.Additional resources
+* xref:assembly-auth-provider-github_proc-configuring-github-app[Configuring a GitHub App in {product-short}]
+* xref:assembly-enabling-authentication_proc-adding-gitlab-oauth-as-an-authentication-provider[Enabling the GitLab OAuth authentication provider]
--- a/modules/templates/ref-creating-templates.adoc
+++ b/modules/templates/ref-creating-templates.adoc
@@ -1,6 +1,6 @@
 // Module included in the following assemblies:
 //
-// * admin/assembly-admin-templates.adoc
+// * assemblies/assembly-admin-templates.adoc
 
 :_mod-docs-content-type: REFERENCE
 [id="ref-creating-templates_{context}"]


modules/templates/proc-adding-templates.adoc
```diff
@@ -0,0 +1,57 @@
...
+* You have created at least one template YAML file in the directory.
+
+.Procedure
+
+. Access the Template Editor by using one of the following options:
++
+image::rhdh/template-editor.png[Template Editor]
+** Open the URL `\https://<rhdh_url>/create/edit` for your {product} instance.
+** Click *Create...* in the navigation menu of the {product} console, then click the overflow menu button and select *Template editor*.
+. Click *Load Template Directory*. This allows you to upload a template from a local directory, by navigating to the directory and clicking *Select*.
```
cmoulliard, line 23: Can you add please a warning to mention that the button `Load Template Director` is available using google Chrome but not with Firefox. Backstage ui help mentions "Only supported in some Chromium-based browsers"

modules/templates/proc-adding-templates.adoc
```diff
@@ -0,0 +1,57 @@
...
+. A *Let site view files?* pop-up appears. Click *View files*.
+. In the *Name ** field, enter a unique name for your template.
```
cmoulliard, line 25: What you describe starting from this item is specific to the template loaded. As you cannot assume what the user will load, then remove such steps: `name`, `Owner`, `Repository Location`

modules/templates/proc-adding-templates.adoc
```diff
@@ -0,0 +1,57 @@
...
+. From the *Owner* drop-down menu, choose an owner for the template.
+. Click *Next*.
+. In the *Repository Location* view, enter information about the the hosted repository to publish the template to:
+.. Select an available *Host* from the drop-down menu.
++
+--
+[NOTE]
```
cmoulliard, line 32: Same remark as before. You assume that the user will load a template containing an action to publish on github

cmoulliard (changes requested):

modules/templates/proc-adding-templates.adoc
abrennan89: If I use gitlab instead of github the requested fields are the same:
image

I think we can assume that an owner and repository location will be required for any repository type. Let's have QE verify this, since I cannot test them all. I think it's better to include this information for completeness and add additional details for other types beyond github and gitlab later if we have more details, rather than leaving this information out entirely.

modules/templates/proc-adding-templates.adoc
abrennan89: See above comment - let's have QE verify whether there are additional details to include here for other repository types. I think from a user perspective it's better to include the information that we have currently at least, since it's the same for github and gitlab.

jayfray12:
I have no issue with the content and seems accurate to me however, is the template editor our recommended approach to templates?  Shouldn't it be storing them in GitHub/GitLab, etc and using the Catalog processors to ingest them to have a single source of truth?

modules/templates/proc-adding-templates.adoc
abrennan89: Good call, I will add something for this

abrennan89:
Reworking per @jayfray12 's suggestion

modules/templates/proc-adding-templates.adoc
```diff
@@ -0,0 +1,45 @@
...
+    - type: url
+      target: https://<repository_url>/example-template.yaml
+   - type: file # <3>
+      target: ./templates # <4>
+   - type: file
+      target: ./templates/example-template.yaml
+# ...
+----
+<1> If you are importing templates from a repository, such as GitHub or GitLab, use the `url` type.
+<2> Specify the URL for either a directory where multiple templates are stored, or an individual template as a YAML file.
```
jayfray12, line 35: Is this a correct statement?
`either a directory where multiple templates are stored`

I didn't know this was possible.  The way I've seen it done is a file that points to other files like it's done here:
https://github.com/janus-idp/software-templates/blob/main/showcase-templates.yaml

modules/templates/proc-adding-templates.adoc
```diff
@@ -0,0 +1,45 @@
...
+<3> If you are importing templates from a local directory, use the `file` type.
+<4> Specify the path to the local directory that contains your template YAML files, or an individual template as a YAML file.
```
jayfray12, line 37: Not sure how specific we want to get here, but if they are running in OpenShift or containers having this locally is trickier and needs to be done via a volume mount

modules/templates/proc-adding-templates.adoc
```diff
@@ -0,0 +1,45 @@
...
+
+. {product} automatically detects new templates that are added to a configured location, including templates that are added to a configured repository directory, and makes them available in the catalog.
```
jayfray12, line 39: This goes with the comment above.  I'm not sure that is the case with the static configuration like above.  The only way a new template would get automatically picked up is if you created that location file like mentioned above and that location file added a new template

modules/templates/proc-adding-templates.adoc
```diff
@@ -0,0 +1,45 @@
...
+
```
jayfray12, line 40: You'll also need to discuss `Catalog Rules` (https://backstage.io/docs/features/software-catalog/configuration#catalog-rules)

By default `Template` is NOT included in the rules so even if you point to a template type backstage won't ingest it unless you add the rule to allow the `Template` type

modules/templates/proc-adding-templates.adoc
abrennan89: Pretty sure I saw docs mentioning this which is why I included it, but I will leave it as individual files for now since I can't find the source atm.

modules/templates/proc-adding-templates.adoc
abrennan89: Should I just remove the local part then?

modules/templates/proc-adding-templates.adoc
cmoulliard: This is correct. You can import a template or a folder containing templates BUT then referenced in a Location YAML file
```
apiVersion: backstage.io/v1alpha1
kind: Location
metadata:
  name: templates
  description: Templates
spec:
  targets:
    - ./quarkus-application/template.yaml
    - ./quarkus-chatbot/template.yaml
    - ./quarkus-rest-client/template.yaml
    - ./quarkus-quickstart/template.yaml
```

modules/templates/proc-adding-templates.adoc
cmoulliard: Yes we should explain how rules apply to templates otherwise you won't see them like also if you don't provide a github token having access to a git repository (= type: url) then backstage will not be able to read and load them

modules/templates/proc-adding-templates.adoc
jayfray12: Right so the missing piece is the `Location` file

modules/templates/proc-adding-templates.adoc
jayfray12: It doesn't hurt to have it here I just don't think it will ever get used

modules/templates/proc-adding-templates.adoc
jayfray12: You could do this automatically like you stated if configure the catalog discovery to look for templates.  This is done using a provider but that is specific to each one (GitHub, GitLab).  Here is a link to GitLab's
https://backstage.io/docs/integrations/gitlab/discovery#configuration

Here is a sample configuration:
```
catalog
  rules:
    - allow: [Template]
  providers:
      templates:
        host: yourgitlabhostname
        entityFilename: template.yaml # This assumes your template files all have the same name of template.yaml
        schedule:
          frequency: { minutes: 30 }
          timeout: { minutes: 3 }
```

modules/templates/proc-adding-templates.adoc
abrennan89: I think that this expands the scope of the original issue enough to require its own Jira as an additional use case.

modules/templates/proc-adding-templates.adoc
abrennan89: Per the above comment, considering the additional configurations involved, I think that automatically picking up templates will need its own Jira as an advanced use case. It feels like quite a large addition to the scope at this point otherwise for something that was originally scoped as an 8 SP Jira.

cmoulliard:
> PR Preview:

Why are you packaging the documentation which is interesting for the developers or platform engineers part of the admin guide ?

abrennan89:
> > PR Preview:
> 
> Why are you packaging the documentation which is interesting for the developers or platform engineers part of the admin guide ?

It didn't seem to me like something that would fit within the installation or getting started guides since it requires modifying config files. Where would you prefer it to be included? Also, it would have been good to have included this feedback in https://github.com/redhat-developer/red-hat-developers-documentation-rhdh/pull/201 instead, because that's where the assembly was actually added. This PR is just expanding on that now existing section.

cmoulliard:
> It didn't seem to me like something that would fit within the installation or getting started guides since it requires modifying config files. Where would you prefer it to be included?

Ideally that should be part of the guide used by the `platform engineer` to configure the backstage idp platform, etc 
We could move it part of the user guide at the condition that we mention clearly the role of the user part of the guide to avoid to have separate documents

Omar-AlJaljuli (approved):
LGTM

modules/templates/proc-creating-templates.adoc
```diff
@@ -1,15 +1,13 @@
 // Module included in the following assemblies:
 //
-// * admin/assembly-admin-templates.adoc
+// * assemblies/assembly-admin-templates.adoc
 
 :_mod-docs-content-type: PROCEDURE
 [id="proc-creating-templates_{context}"]
 = Creating a template by using the Template Editor
 
 You can create a template by using the Template Editor.
```
linfraze, line 9: [Future] This statement is redundant to the heading. It would be nice to add some conceptual information about templates and why a user would want to create one.

modules/templates/proc-adding-templates.adoc
```diff
@@ -0,0 +1,39 @@
...
+You can add an existing template to your {product} instance by using the Catalog Processor.
+
+.Prerequisites
+
+* You have created a directory or repository that contains at least one template YAML file.
+* If you want to use a template that is stored in a repository such as GitHub or GitLab, you must configure a {product} integration for your provider.
+
+.Procedure
+
+* In the `app-config.yaml` configuration file, modify the `catalog.rules` section to include a rule for templates, and configure the `catalog.locations` section to point to the template that you want to add:
```
linfraze, line 18:
```suggestion
* In the `app-config.yaml` configuration file, modify the `catalog.rules` section to include a rule for templates, and configure the `catalog.locations` section to point to the template that you want to add, as shown in the following example:
```

linfraze (approved):
One minor change, otherwise lgtm

modules/templates/proc-creating-templates.adoc
abrennan89: There's some conceptual info in the assembly: https://redhat-developer.github.io/red-hat-developers-documentation-rhdh/pr-219/admin-rhdh/#assembly-admin-templates

This bit is really just to differentiate between different methods of creating templates, since you can do that a couple of ways :)

PR #191, opened 2024-04-20
RHIDP-2109 : GH#190 Docs for operator app-config.yaml configMap

pfeifferj:
#190

PR #228, opened 2024-05-14
RHIDP-2109: Doc procedure for operator app-config.yaml configMap

linfraze:
**Description:**

- Creates an assembly for adding a custom app config to OCP
- Specifies that the existing procedure is for the Helm chart method
- Resolves #191 submitted by @pfeifferj 
- Specifies that the existing procedure in the Admin guide is for the Operator.
- Includes both procedures in the new assembly 

**Versions:**
1.1+

**Jira link:**
https://issues.redhat.com/browse/RHIDP-2109 
https://issues.redhat.com/browse/RHIDP-2365

**Preview link:**
https://redhat-developer.github.io/red-hat-developers-documentation-rhdh/pr-228/getting-started-rhdh/#assembly-add-custom-app-file-openshift_rhdh-getting-started
Sections 3.1, 3.1.1, and 3.1.2

**Reviews:**
 SME: @coreydaley @gazarenkov @rm3l 
 QE: @gustavolira 
 Peer review: TBA


assemblies/assembly-add-custom-app-file-openshift.adoc
```diff
@@ -0,0 +1,24 @@
...
+apiVersion: v1
+metadata:
+  name: app-config-rhdh
+data:
+  app-config-rhdh.yaml: |
+    app:
+      title: Red Hat Developer Hub
+----
+
+You can add the customer application configuration file to {ocp-short} in one of the following ways:
```
gazarenkov, line 18: above mentioned as "custom (not customer) application configuration file".

modules/getting-started/proc-add-custom-app-file-openshift-operator.adoc
```diff
@@ -0,0 +1,33 @@
...
+.Procedure
+
+. From the {ocp-short} web console, select the *ConfigMaps* tab.
+. Click *Create ConfigMap*.
+. From *Create ConfigMap* page, select the *YAML view* option in *Configure via* and make changes to the file, if needed.
+. Click *Create*.
+. Go to the Backstage Custom Resource (CR) instance on the {ocp-short} web console or enter the following command on the command line:
++
+----
+oc edit Backstage rhdh
```
rm3l, line 19: This is just an example, right? Because the CR name might not necessarily be `rhdh`. So maybe add a placeholder and explain that it needs to be replaced with the actual CR name?

assemblies/assembly-add-custom-app-file-openshift.adoc
```diff
@@ -0,0 +1,24 @@
+[id='assembly-add-custom-app-file-openshift_{context}']
+= Adding a custom application configuration file to {ocp-brand-name}
+
+To access the {product}, you must add a custom application configuration file to {ocp-brand-name}. In {ocp-short}, you can use the following content as a base template to create a ConfigMap named `app-config-rhdh`:
```
rm3l, line 4: Quick question about the template for the custom config file. For the operator, we recommended filling in additional fields like the base URL and CORS origin URL. 
So to me, the template would be slightly different for the operator (and a secret would need to be created to store the `BACKEND_SECRET`). Otherwise, the application will not start with the Operator (at least in 1.1), because of a missing `backend.auth.keys.secret`.

See https://access.redhat.com/documentation/en-us/red_hat_developer_hub/1.1/html/administration_guide_for_red_hat_developer_hub/assembly-install-rhdh-ocp#proc-add-custom-app-config-file-ocp_admin-rhdh

```yaml
kind: ConfigMap
apiVersion: v1
metadata:
  name: app-config-rhdh
data:
  "app-config-rhdh.yaml": |
    app:
      title: Red Hat Developer Hub
      baseUrl: https://backstage-<CUSTOM_RESOURCE_NAME>-<NAMESPACE_NAME>.<OPENSHIFT_INGRESS_DOMAIN>
    backend:
      auth:
        keys:
          - secret: "${BACKEND_SECRET}"
      baseUrl: https://backstage-<CUSTOM_RESOURCE_NAME>-<NAMESPACE_NAME>.<OPENSHIFT_INGRESS_DOMAIN>
      cors:
        origin: https://backstage-<CUSTOM_RESOURCE_NAME>-<NAMESPACE_NAME>.<OPENSHIFT_INGRESS_DOMAIN>
```

rm3l: The template in this PR is okay for Helm, as Helm already generates a default app-config file (+ a random backend auth secret), where the base URL and CORS origin URL are constructed from the `global.clusterRouterBase` parameter.

(But I think we can use the template above for both Helm and the Operator).

coreydaley: That should be added to the operator as part of the feature parity between helm <-> operator

rm3l: Sure, let's capture this in the feature parity epic ([RHIDP-2311](https://issues.redhat.com/browse/RHIDP-2311)). But until this is implemented, I think our product docs should reflect the current state.

assemblies/assembly-add-custom-app-file-openshift.adoc
linfraze: Good catch!

modules/getting-started/proc-add-custom-app-file-openshift-operator.adoc
linfraze: Ah, ok. So _rhdh_ should be replaced with a variable in this case? Like _<custom resource name>_?

rm3l: > Ah, ok. So _rhdh_ should be replaced with a variable in this case? Like _< custom resource name >_?

Yes, that's what I was thinking.

assemblies/assembly-add-custom-app-file-openshift.adoc
linfraze: @rm3l is the procedure you linked specific to adding a custom app config file to OCP **using the Operator**? 

If _no,_ then I'll add only the parts of the linked procedure that are applicable to this one. 

If _yes_, then I will take a different approach (like edit the procedure you linked to specify that and then link to it from this doc that we are working on). Ultimately, I see a need to consolidate all of the custom app config file content into one place, rather than leaving it scattered between both the Admin and Getting Started guides, but will address that issue in a future Jira/PR.

linfraze: Note: The linked content is currently under a Helm chart chapter in the Admin guide, but seems that it may not actually be specific to the Helm chart, but rather general or specific to the Operator only.

rm3l: > @rm3l is the procedure you linked specific to adding a custom app config file to OCP **using the Operator**?
> 
> If _no,_ then I'll add only the parts of the linked procedure that are applicable to this one.
> 
> If _yes_, then I will take a different approach (like edit the procedure you linked to specify that it is for the Operator and then link to it from this doc that we are working on).

I would say _yes_, this is specific to the operator at this time. The URLs mentioned in the custom app-config need to be provided explicitly by the user when using the operator, otherwise some parts of the application (like GitHub auth) might not work properly.
For Helm, adding those URLs is not needed explicitly in the custom app-config, as long as users set the correct router hostname (`global.clusterRouterBase`) in their `values` file, as depicted in https://access.redhat.com/documentation/en-us/red_hat_developer_hub/1.1/html/administration_guide_for_red_hat_developer_hub/assembly-install-rhdh-ocp#proc-install-rhdh-ocp-helm_admin-rhdh

> Ultimately, I see a need to consolidate all of the custom app config file content into one place, rather than leaving it scattered between both the Admin and Getting Started guides, but will address that issue in a future Jira/PR.

+1. Definitely agree to have all this consolidated into one place.

gustavolira (approved):

assemblies/assembly-add-custom-app-file-openshift.adoc
```diff
@@ -0,0 +1,25 @@
+[id='assembly-add-custom-app-file-openshift_{context}']
```
abrennan89, line 1:
```suggestion
[id="assembly-add-custom-app-file-openshift_{context}"]
```

assemblies/assembly-add-custom-app-file-openshift.adoc
```diff
@@ -0,0 +1,25 @@
...
+= Adding a custom application configuration file to {ocp-brand-name}
+
+To access the {product}, you must add a custom application configuration file to {ocp-brand-name}. In {ocp-short}, you can use the following content as a base template to create a ConfigMap named `app-config-rhdh`:
+
+[source]
```
abrennan89, line 6:
```suggestion
[source,yaml]
```

assemblies/assembly-add-custom-app-file-openshift.adoc
```diff
@@ -0,0 +1,25 @@
...
+data:
+  app-config-rhdh.yaml: |
+    app:
+      title: {product}
+----
+
+You can add the custom application configuration file to {ocp-short} in one of the following ways:
+
+* The {product} Operator
+* The {product} Helm chart.
```
abrennan89, line 21: I'd update these to "By using The {product} Operator." instead -  it just reads better IMHO

modules/getting-started/proc-add-custom-app-file-openshift-helm.adoc
```diff
@@ -1,19 +1,7 @@
-[id='proc-add-custom-app-file-openshift_{context}']
-= Adding a custom application configuration file to {ocp-brand-name}
+[id='proc-add-custom-app-file-openshift-helm_{context}']
```
abrennan89, line 1:
```suggestion
[id="proc-add-custom-app-file-openshift-helm_{context}"]
```
All the headings should be using " " instead of ' ' per the mod docs guidelines and templates

modules/getting-started/proc-add-custom-app-file-openshift-helm.adoc
```diff
@@ -23,17 +11,15 @@ data:
...
-+
-The list of Helm Releases appears on the page.
-. Click the overflow menu on a Helm release and select *Upgrade*.
-. Use one of the following views to edit the Helm configuration:
+. Go to the *Helm* tab to see the list of Helm releases.
+. Click the overflow menu on the Helm release that you want to use and select *Upgrade*.
+. Use either the *Form view* or *YAML view* to edit the Helm configuration.
 
-** Using *Form view* 
+** Using *Form view*
```
abrennan89, line 20:
```suggestion
** Using *Form view*:
```

modules/getting-started/proc-add-custom-app-file-openshift-helm.adoc
```diff
@@ -43,9 +29,9 @@ The list of Helm Releases appears on the page.
 --
 .. Click *Upgrade*.
 
-** Using *YAML view* 
+** Using *YAML view*
```
abrennan89, line 32:
```suggestion
** Using *YAML view*:
```

modules/getting-started/proc-add-custom-app-file-openshift-operator.adoc
```diff
@@ -0,0 +1,4 @@
+[id='proc-add-custom-app-file-openshift-operator_{context}']
```
abrennan89, line 1:
```suggestion
[id="proc-add-custom-app-file-openshift-operator_{context}"]
```

abrennan89 (commented):
Left some comments, mostly just formatting fixes

abrennan89:
I couldn't find it in this PR, but it also looks like the link in "3.1.2. Adding a custom application configuration file to OpenShift Container Platform using the Operator" is giving a 404 error

modules/getting-started/proc-add-custom-app-file-openshift-helm.adoc
```diff
@@ -23,17 +11,15 @@ data:
 
 . From the {ocp-short} web console, select the *ConfigMaps* tab.
```
abrennan89, line 12: Sort of out of scope for this PR, but when mentioning the OCP web console, I'd recommend saying whether it's in the Developer or Administrator perspective, and also listing admin permissions as a prereq if they're required (usually they are to edit / modify configmaps I think)

modules/getting-started/proc-add-custom-app-file-openshift-helm.adoc
linfraze: Going to leave the colon off for now, since these are headings and not lead-in sentences

modules/getting-started/proc-add-custom-app-file-openshift-helm.adoc
linfraze: That sounds good. I'm working with a SME on a separate PR to improve/reorg all of the _custom app config file_ content and update all the prereqs, etc. Will incorporate this feedback into those changes.

linfraze:
@abrennan89 When I reviewed another PR recently, I noticed that all of the cross-title links in the previews were 404ing but working in prod. Don't know what that's about, but will ensure all links work in Pantheon before publishing.

abrennan89 (approved):
/lgtm

PR #232, opened 2024-05-20
RHIDP-2384: Clean-up Configuring the RHDH CR section & sub-sections

linfraze:
**Versions:**
1.1+

**Description:**
- Deletes the _Configuring the RHDH CR_ section
- Improves the _Adding a custom application configuration file to OCP_ section
- Specifies which procedure is for the Operator and which is for the Helm chart
- Adds both the _Adding a custom application configuration file to OCP using the Operator_ and  _Adding a custom application configuration file to OCP using the Helm chart_ to both the Admin and Getting Started guides using single sourcing best practices 

**Jira link:**
https://issues.redhat.com/browse/RHIDP-2384

**Preview link:**
[Admin guide 2.1.1 & 2.2.1](https://redhat-developer.github.io/red-hat-developers-documentation-rhdh/pr-232/admin-rhdh/#proc-install-rhdh-ocp-operator_admin-rhdh)

[Getting Started guide 3.1.1 & 3.1.2](https://redhat-developer.github.io/red-hat-developers-documentation-rhdh/pr-232/getting-started-rhdh/#ref-rhdh-supported-configs_rhdh-getting-started)

**Reviews:**
SME: @rm3l 
QE: @gustavolira 
Peer review: @abrennan89


modules/installation/proc-config-rhdh-custom-resource.adoc
```diff
@@ -1,4 +1,55 @@
 [id="proc-config-rhdh-custom-resource_{context}"]
 = Configuring the {product-short} Custom Resource
 
-Updates to the Backstage custom resource (CR) are automatically handled by the {product} Operator. However, updates to resources referenced by the CR, such as ConfigMaps or Secrets, are not updated automatically unless the CR itself is updated. If you want to update resources referenced by the CR, then you must manually delete the Backstage Deployment so that the Operator can re-create it with the updated resources.
\ No newline at end of file
+You can change the configuration of the Custom Resource (CR) to program the way that the Operator manages your {product} instance.
```
linfraze, line 4: As a [role], I want to change the configuration of the CR so that I can [???].

modules/installation/proc-config-rhdh-custom-resource.adoc
```diff
@@ -1,4 +1,53 @@
...
+
+[NOTE]
```
rm3l, line 6: I think we can now remove this note. For 1.2 (now that https://issues.redhat.com/browse/RHIDP-1425 (https://github.com/janus-idp/operator/issues/236) is done), this is no longer true. By default, the Operator will automatically restart the RHDH instance upon changes to referenced ConfigMaps/Secrets.
cc @gazarenkov

modules/installation/proc-config-rhdh-custom-resource.adoc
```diff
@@ -1,4 +1,53 @@
...
+====
+Updates to the Backstage CR are automatically handled by the {product} Operator. However, updates to resources referenced by the CR, such as ConfigMaps or Secrets, are not updated automatically unless the CR itself is updated. If you want to update resources referenced by the CR, then you must manually delete the Backstage Deployment so that the Operator can re-create it with the updated resources.
+====
+
+.Prerequisites
+* You have an administrative role in {ocp-brand-name}
```
rm3l, line 12: I don't think users need an admin role in OCP. A regular user with permission to create a ConfigMap and a Secret should be enough.

modules/installation/proc-config-rhdh-custom-resource.adoc
```diff
@@ -1,4 +1,53 @@
...
+* You have added a custom application configuration file to your {ocp-short} instance
```
rm3l, line 13: Not sure I understand this prerequisite. Isn't that what users are trying to achieve with the procedure below?

modules/installation/proc-config-rhdh-custom-resource.adoc
```diff
@@ -1,4 +1,53 @@
...
+----
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: app-config-rhdh
+data:
+  "app-config-rhdh.yaml": |
+    app:
+      title: {product}
+      baseUrl: https://backstage-developer-hub-my-ns.apps.ci-ln-vtkzr22-72292.origin-ci-int-gce.dev.rhcloud.com
```
rm3l, line 31:
```suggestion
      baseUrl: https://backstage-<CUSTOM_RESOURCE_NAME>-<NAMESPACE_NAME>.<OPENSHIFT_INGRESS_DOMAIN>
```

The whole YAML is a template, right? So maybe we should not hardcode values there..

modules/installation/proc-config-rhdh-custom-resource.adoc
```diff
@@ -1,4 +1,53 @@
...
+    backend:
+      auth:
+        keys:
+          - secret: "${BACKEND_SECRET}"
+      baseUrl: https://backstage-backstage-sample-my-ns.apps.ci-ln-vtkzr22-72292.origin-ci-int-gce.dev.rhcloud.com
```
rm3l, line 36:
```suggestion
      baseUrl: https://backstage-<CUSTOM_RESOURCE_NAME>-<NAMESPACE_NAME>.<OPENSHIFT_INGRESS_DOMAIN>
```

modules/installation/proc-config-rhdh-custom-resource.adoc
```diff
@@ -1,4 +1,53 @@
...
+      cors:
+        origin: https://backstage-backstage-sample-my-ns.apps.ci-ln-vtkzr22-72292.origin-ci-int-gce.dev.rhcloud.com
```
rm3l, line 38:
```suggestion
        origin: https://backstage-<CUSTOM_RESOURCE_NAME>-<NAMESPACE_NAME>.<OPENSHIFT_INGRESS_DOMAIN>
```

modules/installation/proc-config-rhdh-custom-resource.adoc
```diff
@@ -1,4 +1,53 @@
...
+
+.Prerequisites
+* You have an administrative role in {ocp-brand-name}
+* You have added a custom application configuration file to your {ocp-short} instance
+
+.Procedure
+
+To configure your {product-short} CR file for your {product-short} instance, complete the following steps:
+
+. Add a custom application configuration file to {ocp-short} and reference it in the CR. In {ocp-short}, you can use the following example as a base template to create a ConfigMap, such as `app-config-rhdh.yaml`:
```
rm3l, line 19: Where is explained how to reference the custom configuration file in the CR?

EDIT: I can see it in "Adding a custom application configuration file to OpenShift Container Platform using the Operator" (Step 3), but the text (`Select the YAML view option in Configure via and make the changes to the file, if necessary.`) should instruct to copy and adjust based on the content of the ConfigMap template below, IMO.

modules/installation/proc-config-rhdh-custom-resource.adoc
linfraze: @rm3l  Based on our discussion in Slack, I thought we determined that it makes more sense to add the custom app config file to OCP BEFORE configuring the CR. I still plan to change the order in which these two steps are presented in the doc (unless you advise against that), but we can remove this as a preq either way, if it is not actually required / highly recommended that this step occur first.

rm3l: > Based on our discussion in Slack, I thought we determined that it makes more sense to add the custom app config file to OCP BEFORE configuring the CR.

Ah yes, you're right! I found this prereq was conflicting with step 1 ("Add a custom application configuration file to OCP") of the procedure, and that's why I was confused.
The order we discussed (custom app config file first, then configuring the CR) makes sense.

> I still plan to change the order in which these two steps are presented in the doc so that Adding the custom app config file occurs before Configuring CR (unless you advise against that)

+1 for this order.

> but we can remove this as a prereq either way, if it is not actually required / highly recommended that this step occur first.

Is this section only about the initial deployment of RHDH? Or also further customization of an existing RHDH instance? Or both?
If we are talking about the initial deployment of RHDH, there won't be any existing custom application configuration file. Even for a customization of an existing instance, even if the custom application configuration file exists, its content might be as expected already.
So I would suggest removing this as prereq. WDYT?

modules/installation/proc-config-rhdh-custom-resource.adoc
linfraze: @rm3l this is an intro to explain what this procedure is and why a user might want or need to do this procedure. Can you help me fill in the user story? 

As a [role], I want to change the configuration of the CR so that I can [???].

linfraze: Ah, good to know, thank you! 

Note: I formatted this text as a Note, but it's currently the entirely of the _Configuring the RHDH CR_ section published in the Admin guide (https://docs.redhat.com/en/documentation/red_hat_developer_hub/1.1/html/administration_guide_for_red_hat_developer_hub/assembly-install-rhdh-ocp#proc-config-rhdh-custom-resource_admin-rhdh). So it's good that we're repurposing that section with new (relocated) content.

modules/installation/proc-config-rhdh-custom-resource.adoc
linfraze: Based on the existing content published in the Admin guide, it sounds like the first step to _Configuring the RHDH CR_ is adding a custom app config file to OCP. Is that true? If so, since Adding a custom app config file to OCP is a separate procedure, it seems like we should tell users to do that procedure first and link to it in the Prerequisites section. 

Also we currently have "Add a custom app config file to OCP..." as the first step of the _Configuring the RHDH CR_ procedure, but it probably shouldn't be. It should probably just be a Prereq instead. 

But I have to admit that I still lack a lot of technical knowledge about this whole process and might be missing or misunderstanding some key or complex parts.

linfraze: PS - In a future PR, I'm also planning to move these sections out from under the Deploying RHDH on OCP section so that they come first in the user journey before the user even tries to Deploy anything.

rm3l: > Based on the existing content published in the Admin guide, it sounds like the first step to _Configuring the RHDH CR_ is adding a custom app config file to OCP so you can reference it in the CR file. Is that true? If so, since Adding a custom app config file to OCP is a separate procedure, it seems like we should tell users to do that procedure first and link to it in the Prerequisites section.
> 
> Also we currently have "Add a custom app config file to OCP..." as the first step of the _Configuring the RHDH CR_ procedure, but it probably shouldn't be. It should probably just be a Prereq instead. And step 1 should by solely focused on editing the CR to reference the custom app config file that must be created before you can even do this procedure. WDYT?
> 

Yes, that makes sense. Users need to create the custom ConfigMap (and Secret too) first, then reference them in the CR, even for the very first deployment of RHDH.

rm3l (commented):
A few comments, but overall this is much better. Thanks @linfraze !!

modules/installation/proc-install-rhdh-ocp-operator.adoc
```diff
@@ -19,7 +19,7 @@ For more information about creating a project in {ocp-short}, see link:https://d
 . From the *Developer Catalog* panel, click *Operator Backed*.
 . In the *Filter by keyword* box, enter _{product-short}_ and click the *{product}* card.
 . Click *Create*.
-. Add <<Configuring the {product-short} Custom Resource, custom configurations>> for the {product} instance.
+. Add <<Adding a custom application configuration file to {ocp-short} using the Operator, custom configurations>> for the {product} instance.
```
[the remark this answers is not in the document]
linfraze: Yes, that is great feedback. However, the only reason I am editing the installation procedure at this time, though, is to fix the xref issue that I created by deleting the _Configuring the {product-short} Custom Resource_ file from the repo, which was being referenced by the installation procedure. We will fix all of the installation content when we create the new installation guide. I will add a task to the RHIDP-1804 Epic that addresses this issue.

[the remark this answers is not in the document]
linfraze: https://issues.redhat.com/browse/RHIDP-2503

rm3l (approved):

gustavolira (approved):

modules/installation/proc-add-custom-app-config-file-ocp-operator.adoc
```diff
@@ -0,0 +1,95 @@
+[id="proc-add-custom-app-config-file-ocp_{context}"]
```
abrennan89, line 1:
```suggestion
[id="proc-add-custom-app-config-file-ocp-operator_{context}"]
```
ID needs to match the filename

modules/installation/proc-add-custom-app-config-file-ocp-operator.adoc
```diff
@@ -0,0 +1,95 @@
...
+= Adding a custom application configuration file to {ocp-short} using the Operator
+
+A custom application configuration file is a ConfigMap that you can use to change the configuration of your {product} instance to meet the needs and preferences of your organization. If you are deploying your {product-short} instance on {ocp-brand-name}, you can use the {product} Operator to add a custom application configuration file to your {ocp-short} instance by creating the ConfigMap and referencing it in the {product-short} Custom Resource.
```
abrennan89, line 4:
```suggestion
A custom application configuration file is a `ConfigMap` object that you can use to change the configuration of your {product} instance. If you are deploying your {product-short} instance on {ocp-brand-name}, you can use the {product} Operator to add a custom application configuration file to your {ocp-short} instance, by creating the `ConfigMap` object and referencing it in the {product-short} custom resource (CR).
```
Couple of things:
- Suggested editing for minimalism to remove fluff in the first sentence.
- Suggested formatting of config map. Per the OCP guidelines and the RH supplementary style guide, for consistency across the org, we should either use the API object formatting if we're referring to a specific object ("`ConfigMap` object"), or it should be formatted as "config map".
- Suggested formatting of custom resource. Per the OCP guidelines and RHSG we should use "custom resource (CR)" for the first instance, and then you can use CR throughout the rest of the module etc if you like.

Links for ref:
- https://redhat-documentation.github.io/supplementary-style-guide/#custom-resource
- https://redhat-documentation.github.io/supplementary-style-guide/#config-map
- https://github.com/openshift/openshift-docs/blob/main/contributing_to_docs/doc_guidelines.adoc#api-object-formatting

modules/installation/proc-add-custom-app-config-file-ocp-operator.adoc
```diff
@@ -0,0 +1,95 @@
...
+
+The custom application configuration file contains a sensitive environment variable, named `BACKEND_SECRET`, which is a mandatory backend authentication key that {product-short} uses to reference an environment variable defined in an {ocp-short} Secret. You must create a custom Secret, named 'secrets-rhdh', and reference it in the {product-short} Custom Resource.
```
abrennan89, line 6:
```suggestion
The custom application configuration file contains a sensitive environment variable named `BACKEND_SECRET`. This variable contains a mandatory backend authentication key that {product-short} uses to reference an environment variable defined in an {ocp-short} secret. You must create a secret named 'secrets-rhdh', and reference it in the {product-short} CR.
```

Suggestion to break up sentences so that this is easier to parse.
Similar to above comment, unless referring to a specific `Secret` object, this should be lowercase "secret" for OCP references.

modules/installation/proc-add-custom-app-config-file-ocp-operator.adoc
```diff
@@ -0,0 +1,95 @@
...
+
+[NOTE]
+====
+You are responsible for protecting your {product} installation from external and unauthorized access. Manage the backend authentication key like any other secret. Meet strong password requirements, do not expose it in any configuration files, and only inject it into configuration files as an environment variable.
+====
+
+.Prerequisites
+* You have an active {ocp-brand-name} account.
+* Your administrator has installed the {product} Operator in {ocp-short}. For more information, see link:{LinkAdminGuide}[Installing the {product} Operator].
+* You have created the {product} Custom Resource in {ocp-short}.
```
abrennan89, line 16:
```suggestion
* You have created the {product} CR in {ocp-short}.
```
Can we clarify which CR we're referring to here? Maybe throughout this doc rather than referring to it as the {product} CR, we can mention it by name and link to docs about creating it? Presumably it's maybe the `app-config.yaml`, but I'm not sure and it's not really clear?

modules/installation/proc-add-custom-app-config-file-ocp-operator.adoc
```diff
@@ -0,0 +1,95 @@
...
+
+.Procedure
+. From the *Developer* perspective, select the *ConfigMaps* tab.
```
abrennan89, line 19:
```suggestion
. From the *Developer* perspective in the {ocp-short} web console, select the *ConfigMaps* tab.
```
Should this be "view" instead of "tab", similar to Topology?

modules/installation/proc-add-custom-app-config-file-ocp-operator.adoc
```diff
@@ -0,0 +1,95 @@
...
+      title: {product}
+      baseUrl: https://backstage-<CUSTOM_RESOURCE_NAME>-<NAMESPACE_NAME>.<OPENSHIFT_INGRESS_DOMAIN>
+    backend:
+      auth:
+        keys:
+          - secret: "${BACKEND_SECRET}"
+      baseUrl: https://backstage-<CUSTOM_RESOURCE_NAME>-<NAMESPACE_NAME>.<OPENSHIFT_INGRESS_DOMAIN>
+      cors:
+        origin: https://backstage-<CUSTOM_RESOURCE_NAME>-<NAMESPACE_NAME>.<OPENSHIFT_INGRESS_DOMAIN>
+----
```
abrennan89, line 41: Per the Red Hat supplementary style guide, user-replaceable values should use `<this_format>`, so these should be lowercase throughout the doc. See https://redhat-documentation.github.io/supplementary-style-guide/#user-replaced-values

modules/installation/proc-add-custom-app-config-file-ocp-operator.adoc
```diff
@@ -0,0 +1,95 @@
...
++
+.. Use the mandatory backend authentication key for {product} to reference an environment variable defined in an {ocp-short} Secret.
```
abrennan89, line 43:
```suggestion
.. Use the mandatory backend authentication key for {product} to reference an environment variable defined in an {ocp-short} secret.
```

modules/installation/proc-add-custom-app-config-file-ocp-operator.adoc
```diff
@@ -0,0 +1,95 @@
...
+.. Set the external URL of your {product} instance in the `app.baseUrl`, `backend.baseUrl` and `backend.cors.origin` fields of the application configuration. By default, the URL is similar to the following example:
```
abrennan89, line 44: Same comment as above re https://redhat-documentation.github.io/supplementary-style-guide/#user-replaced-values being lowercase

modules/installation/proc-add-custom-app-config-file-ocp-operator.adoc
```diff
@@ -0,0 +1,95 @@
...
+`pass:c[https://backstage-<CUSTOM_RESOURCE_NAME>-<NAMESPACE_NAME>.<OPENSHIFT_INGRESS_DOMAIN>;]`.
++
+[NOTE]
+====
+You can use the `oc get ingresses.config/cluster -o jsonpath='{.spec.domain}'` command to display your ingress domain. If you want to use a different host or sub-domain, customize the Custom Resource `spec.application.route` field and adjust the application configuration accordingly.
```
abrennan89, line 49:
```suggestion
You can use the `oc get ingresses.config/cluster -o jsonpath='{.spec.domain}'` command to display your ingress domain. If you want to use a different host or sub-domain, customize the CR `spec.application.route` field and adjust the application configuration accordingly.
```

modules/installation/proc-add-custom-app-config-file-ocp-operator.adoc
```diff
@@ -0,0 +1,95 @@
...
+====
+
+. Click *Create*.
+. Select the *Secrets* tab.
```
abrennan89, line 53: Same for all instances of tab - should these be "view" if they're clicked on from the same left nav as the Topology view?

modules/installation/proc-add-custom-app-config-file-ocp-operator.adoc
```diff
@@ -0,0 +1,95 @@
...
+. Click *Create Key/value Secret*.
+. Create a custom secret named `secrets-rhdh`.
```
abrennan89, line 55:
```suggestion
. Create a secret named `secrets-rhdh`.
```
I think "custom" is superfluous throughout if they're creating a new secret.

modules/installation/proc-add-custom-app-config-file-ocp-operator.adoc
```diff
@@ -0,0 +1,95 @@
...
+      secrets:
+         - name: secrets-rhdh
+    extraFiles:
+      mountPath: /opt/app-root/src
+    replicas: 1
+    route:
+      enabled: true
+  database:
+    enableLocalDb: true
+----
```
abrennan89, line 89: In this YAML example, I would include the API version, kind, etc so that it's clear what the custom resource is that's being modified here. For example, if it was a config map:
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: example
spec:....
```

I think it's good practice to always include this for context for users.

modules/installation/proc-add-custom-app-config-file-ocp-operator.adoc
```diff
@@ -0,0 +1,95 @@
...
+node -p 'require("crypto").randomBytes(24).toString("base64")'
+----
+
+. Click *Create*.
+. Select the *Topology* view.
+. Click the overflow menu for the {product} instance that you want to use and select *Edit Backstage* to load the YAML view of the {product} instance.
++
+image::rhdh/operator-install-2.png[]
+
+. In the Custom Resource file, enter the name of the custom application configuration ConfigMap as the value for the `spec.application.appConfig.configMaps` field and the name of the custom Secret as the value for the `spec.application.extraEnvs.secrets` field. For example:
```
abrennan89, line 69:
```suggestion
. In the CR, enter the name of the custom application configuration config map as the value for the `spec.application.appConfig.configMaps` field, and enter the name of your secret as the value for the `spec.application.extraEnvs.secrets` field. For example:
```

modules/installation/proc-install-rhdh-ocp-operator.adoc
abrennan89, line 22: Couldn't find this in the preview to check the link.

abrennan89:
FYI, just a tip in case you're not aware @linfraze - if you squash your commits before you merge this, it will make your life easier if you need to do a manual cherrypick, since you'll only need to CP one commit 🙏🏻

modules/installation/proc-install-rhdh-ocp-operator.adoc
linfraze: This is in the Getting Started guide, I think. Just a little xref fix since we deleted the CR config content that was being referenced here.

modules/installation/proc-add-custom-app-config-file-ocp-operator.adoc
```diff
@@ -0,0 +1,99 @@
...
+
+.Prerequisites
+* You have an active {ocp-brand-name} account.
+* Your administrator has installed the {product} Operator in {ocp-short}. For more information, see link:{LinkAdminGuide}[Installing the {product} Operator].
+* You have created the {product} CR in {ocp-short}.
+
+.Procedure
+. From the *Developer* perspective in the {ocp-short} web console, select the *ConfigMaps* view.
+. Click *Create ConfigMap*.
+. Select the *YAML view* option in *Configure via* and use the following example as a base template to create a ConfigMap, such as `app-config-rhdh.yaml`:
```
abrennan89, line 21:
```suggestion
. Select the *YAML view* option in *Configure via* and use the following example as a base template to create a `ConfigMap` object, such as `app-config-rhdh.yaml`:
```

abrennan89 (approved):
One last minor formatting thing, otherwise lgtm, great work!

abrennan89:
@linfraze does this need to be cherrypicked to 1.1 also?

PR #247, opened 2024-05-31
RHIDP-2062: Align text and code snippet

abrennan89:
**Versions:**
1.1+

**Jira link:**
- https://issues.redhat.com/browse/RHIDP-2062

**Preview link:**
https://redhat-developer.github.io/red-hat-developers-documentation-rhdh/pr-247/admin-rhdh/#proc-install-rhdh-airgapped-environment-ocp-helm_admin-rhdh

**Reviews:**
- [x] SME: @coreydaley @nickboldt 
- QE: N/A, small change to a command TBC by SME
- [x] Peer review: @linfraze

### DIFF
--- a/modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
+++ b/modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
@@ -73,7 +73,12 @@ podman pull registry.redhat.io/rhdh/rhdh-hub-rhel9:{product-chart-version}
 podman pull registry.redhat.io/rhel9/postgresql-15:latest
 ----
 
-. Push both images to the internal {ocp-short} image registry by running the following command:
+. Push both images to the internal {ocp-short} image registry by running the following commands:
++
+[source,terminal]
+----
+podman push --remove-signatures registry.redhat.io/rhdh/rhdh-hub-rhel9:{product-chart-version} default-route-openshift-image-registry.<hostname>/<project_name>/rhdh-hub-rhel9:{product-chart-version}
+----
 +
 [source,terminal]
 ----


modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -73,7 +73,12 @@ podman pull registry.redhat.io/rhdh/rhdh-hub-rhel9:{product-chart-version}
 podman pull registry.redhat.io/rhel9/postgresql-15:latest
 ----
 
-. Push both images to the internal {ocp-short} image registry by running the following command:
+. Push both images to the internal {ocp-short} image registry by running the following commands:
```
coreydaley, line 76: I only see one command listed? Should `commands` be singular instead of plural?

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -73,7 +73,12 @@ podman pull registry.redhat.io/rhdh/rhdh-hub-rhel9:{product-chart-version}
...
++
+[source,terminal]
+----
+podman push --remove-signatures registry.redhat.io/rhdh/rhdh-hub-rhel9:{product-chart-version} default-route-openshift-image-registry.<hostname>/<project_name>/rhdh-hub-rhel9:{product-chart-version}
+----
```
nickboldt, line 81: why do these commands need to be in seprate code blocks? just means more steps for a user to copy-and-paste TWICE instead of once.

abrennan89: Mostly just a good habit - it comes from working on [OpenShift docs](https://github.com/openshift/openshift-docs/blob/main/contributing_to_docs/doc_guidelines.adoc#single-command-per-code-block).

Now that [docs.redhat.com](https://docs.redhat.com/en/documentation/red_hat_developer_hub/1.1/html/getting_started_with_red_hat_developer_hub/index) is up and running, there is a copy to clipboard functionality (although we didn't have it AFAIK on our old customer portal access.redhat.com docs), but I think given the docs.redhat.com docs it makes good sense to follow this convention to avoid breaking anything.

image

nickboldt: so it's "make people do two steps only because that's the convention, and we don't want to make things easier." OK, I guess?

linfraze (approved):
/lgtm

nickboldt:
@abrennan89 should this also be picked to 1.2.x?

PR #263, opened 2024-06-06
RHIDP-1582: Add docs for enabling metrics collection

abrennan89:
**Versions:**
1.1+ (1.0 docs are no longer maintained)

**Jira link:**
- https://issues.redhat.com/browse/RHIDP-1582

**Preview link:**
- https://redhat-developer.github.io/red-hat-developers-documentation-rhdh/pr-263/admin-rhdh/#assembly-rhdh-observability

**Reviews:**
- [x] SME: @rm3l 
- [x] QE: @Omar-AlJaljuli
- [x] Peer review: @linfraze


modules/observability/proc-admin-enabling-metrics-ocp-operator.adoc
```diff
@@ -0,0 +1,69 @@
...
+  labels:
+    app.kubernetes.io/instance: <custom_resource_name>
+    app.kubernetes.io/name: backstage
+spec:
+  namespaceSelector:
+    matchNames:
+      - <project_name>
+  selector:
+    matchLabels:
+      janus-idp.io/app: backstage-<custom_resource_name>
```
rm3l, line 52:
```suggestion
      rhdh.redhat.com/app: backstage-<custom_resource_name>
```

The label name has changed in 1.1, and we forgot to update the upstream doc. I'll create a follow-up issue to fix this.

modules/observability/proc-admin-enabling-metrics-ocp-operator.adoc
```diff
@@ -0,0 +1,69 @@
...
+  endpoints:
+  - port: http-backend
```
rm3l, line 54:
```suggestion
  - port: backend
```
Same point for the port name, which is not correct: https://github.com/janus-idp/operator/blob/1.1.x/config/manager/default-config/deployment.yaml#L88 in 1.1 and https://github.com/janus-idp/operator/blob/1.2.x/config/manager/default-config/deployment.yaml#L97 in 1.2

modules/observability/proc-admin-enabling-metrics-ocp-operator.adoc
```diff
@@ -0,0 +1,69 @@
...
+
+.Procedure
+
+Currently, the {product} Operator does not support creating a `ServiceMonitor` custom resource (CR) by default. You must complete the following steps to create a `ServiceMonitor` custom resource (CR) to scrape metrics from the endpoint.
+
+. Substitute the values for your environment and then run the following commands to create the `ServiceMonitor` CR:
++
+[source,terminal]
+----
+<custom_resource_name>=my-rhdh
```
rm3l, line 27: From a user perspective, I think it is a bit unclear if they need to run this as a command or not. Instead of an example value like `my-rhdh`, could we maybe just describe the meaning of this parameter?
For example:
> Replace `<custom_resource_name>` with the name of your {product} Custom Resource.
> Replace `<project_name>` with the name of the {ocp-short} project where your {product} instance is running.

modules/observability/proc-admin-enabling-metrics-ocp-operator.adoc
```diff
@@ -0,0 +1,69 @@
...
+----
++
+[source,terminal]
+----
+<project_name>=my-project
+----
++
+[source,yaml]
+----
+cat <<EOF > /tmp/$<custom_resource_name>.ServiceMonitor.yaml
```
rm3l, line 37: > cat <<EOF > /tmp/$<custom_resource_name>.ServiceMonitor.yaml

The goal with the upstream doc was to provide a way for people to just replace the value of `CR_NAME` and `MY_PROJECT`, then execute the `cat <<EOF ...` command as is right away from their terminal. The shell would then automatically substitute `${CR_NAME}` and `${MY_PROJECT}` accordingly.

Here, it looks like we are expecting users to replace all occurrences of those variables. So IMO, the `cat` command is not required and we can just rewrite this procedure like so:
1. manually create a YAML file named `developer-hub.ServiceMonitor.yaml` (with the content being https://github.com/abrennan89/red-hat-developers-documentation-rhdh/blob/RHIDP-1582/modules/observability/proc-admin-enabling-metrics-ocp-operator.adoc?plain=1#L38-L55, and the variables substituted)
2. Run the `oc apply -f developer-hub.ServiceMonitor.yaml` command

WDYT?

rm3l (changes requested):

modules/observability/proc-admin-enabling-metrics-ocp-operator.adoc
```diff
@@ -0,0 +1,69 @@
...
+
+You can enable and view metrics for an Operator-installed {product} instance from the *Developer* perspective of the {ocp-short} web console.
+
+.Prerequisites
+
+* Your {ocp-short} cluster has link:https://docs.openshift.com/container-platform/latest/observability/monitoring/enabling-monitoring-for-user-defined-projects.html[monitoring for user-defined projects] enabled.
+
+* You have installed {product} on {ocp-short} using the {product} Operator.
+
+* You have installed the {ocp-short} `oc` CLI.
```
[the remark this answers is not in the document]
abrennan89: Updated this to use the same attribute as the OCP docs, good catch

modules/observability/proc-admin-enabling-metrics-ocp-operator.adoc
abrennan89: I had actually wondered if it made sense to document the process this way :) I wasn't sure if there was a reason why you set the variables and then ran the cat command, but I think it makes more sense and is more consistent with other RH docs to document it by just telling users to create the YAML CR and then apply it :)

assemblies/assembly-rhdh-observability.adoc
```diff
@@ -0,0 +1,12 @@
+:_mod-docs-content-type: ASSEMBLY
+[id="assembly-rhdh-observability"]
+= Enabling observability for {product}
+:context: assembly-rhdh-observability
+
+toc::[]
+
+// Metrics
+{product} provides a `/metrics` endpoint that can be scraped by the Prometheus collector. You can use Prometheus and Grafana to monitor your {product} instance and observe metrics that are scraped from this endpoint.
```
rm3l, line 9:
```suggestion
```

Do we need to mention Prometheus and Grafana in our product docs which are dedicated to OCP? We are documenting observability via the `ServiceMonitor` construct, not Prometheus and Grafana directly.

modules/observability/proc-admin-enabling-metrics-ocp-operator.adoc
```diff
@@ -0,0 +1,57 @@
...
++
+[source,yaml]
+----
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  name: <custom_resource_name> # <1>
+  namespace: <project_name> # <2>
+  labels:
+    app.kubernetes.io/instance: <custom_resource_name>
```
rm3l, line 31:
```suggestion
    app.kubernetes.io/instance: <custom_resource_name> # <1>
```

modules/observability/proc-admin-enabling-metrics-ocp-operator.adoc
```diff
@@ -0,0 +1,57 @@
...
+    app.kubernetes.io/name: backstage
+spec:
+  namespaceSelector:
+    matchNames:
+      - <project_name>
+  selector:
+    matchLabels:
+      rhdh.redhat.com/app: backstage-<custom_resource_name>
```
rm3l, line 39:
```suggestion
      rhdh.redhat.com/app: backstage-<custom_resource_name> # <1>
```

modules/observability/proc-admin-enabling-metrics-ocp-operator.adoc
```diff
@@ -0,0 +1,57 @@
...
+metadata:
+  name: <custom_resource_name> # <1>
+  namespace: <project_name> # <2>
+  labels:
+    app.kubernetes.io/instance: <custom_resource_name>
+    app.kubernetes.io/name: backstage
+spec:
+  namespaceSelector:
+    matchNames:
+      - <project_name>
```
rm3l, line 36:
```suggestion
      - <project_name> # <2>
```

rm3l (commented):
One last question: do we need to repeat the `callouts` where they are used or not?

modules/observability/proc-admin-enabling-metrics-ocp-operator.adoc
abrennan89: I think including the callouts once is enough - using the same number multiple times can mess with the formatting, and otherwise we'd just be repeating the same callouts with different numbers

assemblies/assembly-rhdh-observability.adoc
abrennan89: Updated this in line with what's in the OCP docs and added a link

rm3l (approved):

Omar-AlJaljuli:
There is currently an issue where if you set up your helm through the CLI, you wouldn't be able to upgrade it through the console. Would it be possible to also add a subsection that covers upgrading the helm release through the CLI?

rm3l:
> There is currently an issue where if you set up your helm through the CLI, you wouldn't be able to upgrade it through the console. Would it be possible to also add a subsection that covers upgrading the helm release through the CLI?

But this is out of the scope of this PR, no? I think this would be a great suggestion for [RHIDP-2614](https://issues.redhat.com/browse/RHIDP-2614) (#287) maybe?

Omar-AlJaljuli:
> > There is currently an issue where if you set up your helm through the CLI, you wouldn't be able to upgrade it through the console. Would it be possible to also add a subsection that covers upgrading the helm release through the CLI?
> 
> But this is out of the scope of this PR, no? I think this would be a great suggestion for [RHIDP-2614](https://issues.redhat.com/browse/RHIDP-2614) (#287) maybe?

That makes sense, I guess that it's an overarching issue that isn't specific to this pr.

Omar-AlJaljuli (approved):
LGTM

modules/observability/proc-admin-enabling-metrics-ocp-operator.adoc
```diff
@@ -0,0 +1,57 @@
...
+
+.Prerequisites
+
+* Your {ocp-short} cluster has link:https://docs.openshift.com/container-platform/latest/observability/monitoring/enabling-monitoring-for-user-defined-projects.html[monitoring for user-defined projects] enabled.
+* You have installed {product} on {ocp-short} using the {product} Operator.
+* You have installed the {oc-first}.
+
+.Procedure
+
+Currently, the {product} Operator does not support creating a `ServiceMonitor` custom resource (CR) by default. You must complete the following steps to create a `ServiceMonitor` custom resource (CR) to scrape metrics from the endpoint.
```
linfraze, line 19:
```suggestion
Currently, the {product} Operator does not support creating a `ServiceMonitor` custom resource (CR) by default. You must complete the following steps to create a `ServiceMonitor` CR to scrape metrics from the endpoint.
```

linfraze (approved):
/lgtm

assemblies/assembly-rhdh-observability.adoc
```diff
@@ -0,0 +1,17 @@
+:_mod-docs-content-type: ASSEMBLY
+[id="assembly-rhdh-observability"]
+= Enabling observability for {product} on {ocp-short}
+:context: assembly-rhdh-observability
+
+toc::[]
+
+// Metrics
+In {ocp-short}, metrics are exposed through an HTTP service endpoint under the `/metrics` canonical name. You can create a `ServiceMonitor` custom resource (CR) to scrape metrics from a service endpoint in a user-defined project.
```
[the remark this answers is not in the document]
abrennan89: It is indeed common / an industry term to metrics / Prometheus :)

modules/observability/proc-admin-enabling-metrics-ocp-helm.adoc
```diff
@@ -0,0 +1,43 @@
...
+
+:_mod-docs-content-type: PROCEDURE
+[id="proc-admin-enabling-metrics-ocp-helm_{context}"]
+= Enabling metrics monitoring in a Helm chart installation on an {ocp-short} cluster
+
+You can enable and view metrics for a {product} Helm deployment from the *Developer* perspective of the {ocp-short} web console.
+
+.Prerequisites
+
+* Your {ocp-short} cluster has link:https://docs.openshift.com/container-platform/latest/observability/monitoring/enabling-monitoring-for-user-defined-projects.html[monitoring for user-defined projects] enabled.
```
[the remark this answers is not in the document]
abrennan89: I will leave it as is, because there's no way to know if the person is an admin or not :) so it's easier to just say it should be enabled I think

PR #347, opened 2024-06-27
[RHIDP-2941]: Remove italics from code block

abrennan89:
**Version(s):**
1.2+

**Issue:**
https://issues.redhat.com/browse/RHIDP-2941

**Link to docs preview:**
- https://redhat-developer.github.io/red-hat-developers-documentation-rhdh/pr-347/admin-rhdh/#proc-rhdh-deploy-eks-using-helm_assembly-rhdh-integration-aws
- https://redhat-developer.github.io/red-hat-developers-documentation-rhdh/pr-347/admin-rhdh/#proc-rhdh-deploy-aks-helm_assembly-rhdh-integration-aks

**Reviews:**
- SME: N/A
- QE: N/A
- [x] Docs review: @linfraze

### DIFF
--- a/modules/admin/proc-rhdh-deploy-aks-helm.adoc
+++ b/modules/admin/proc-rhdh-deploy-aks-helm.adoc
@@ -113,8 +113,8 @@ helm repo add openshift-helm-charts https://charts.openshift.io/
 +
 [source,terminal,subs="quotes+"]
 ----
-DEPLOYMENT_NAME=_<redhat-developer-hub>_
-NAMESPACE=_<rhdh>_
+DEPLOYMENT_NAME=<redhat-developer-hub>
+NAMESPACE=<rhdh>
 kubectl create namespace ${NAMESPACE}
 kubectl config set-context --current --namespace=${NAMESPACE}
 ----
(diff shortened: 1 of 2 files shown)


modules/admin/proc-rhdh-deploy-aks-helm.adoc
```diff
@@ -113,8 +113,8 @@ helm repo add openshift-helm-charts https://charts.openshift.io/
 +
 [source,terminal,subs="quotes+"]
 ----
-DEPLOYMENT_NAME=_<redhat-developer-hub>_
-NAMESPACE=_<rhdh>_
+DEPLOYMENT_NAME=<redhat-developer-hub>
```
linfraze, line 116: A SME pointed out that we are using examples (`redhat-developer-hub`) in some places and variables (`<deployment_name>`) in others and recommended that we consistently use variables. Is this a case of mixing the two so that an example is formatted like a variable? 

Idk what is necessarily right/best or if we should change it. Might also be a bit out of scope for this PR if all you're doing is removing the italics. But WDYT?

linfraze (approved):
One Q but the change lgtm

modules/admin/proc-rhdh-deploy-aks-helm.adoc
abrennan89: I would say yes it's out of scope for this PR, however the reason why it's mixed here is because they're setting it as a variable at the start of the section so that the commands after can be copy and pasted, using `$DEPLOYMENT_NAME` instead of typing out the name each time. I do not really agree with this approach, since it's adding an additional step for users who may not understand or want to do this, without explanation, but for now I am going to leave it as is.

PR #369, opened 2024-07-11
RHIDP-3160: Update links in Helm CLI doc

linfraze:
**Version(s):**
1.2 

**Issue:**
https://issues.redhat.com/browse/RHIDP-3160

**Link to docs preview:**
https://redhat-developer.github.io/red-hat-developers-documentation-rhdh/pr-369/admin-rhdh/#proc-installing-rhdh-on-openshift-with-helm-cli_assembly-installing-rhdh-on-ocp-by-using-the-helm-chart

**Reviews:**
- [ ] SME: @nickboldt (optional)
- [ ] QE: N/A
- [x] Docs: @Preeticp

### DIFF
--- a/modules/installation/proc-installing-rhdh-on-openshift-with-helm-cli.adoc
+++ b/modules/installation/proc-installing-rhdh-on-openshift-with-helm-cli.adoc
@@ -5,10 +5,10 @@ You can use the Helm CLI to install {product} on {ocp-brand-name}.
 
 .Prerequisites
 * You have installed the {openshift-cli} on your workstation.
-* You are logged in to your OpenShift Container Platform account.
-* A user with the OpenShift Container Platform admin role has configured the appropriate roles and permissions within your project to create an application. For more information about OpenShift Container Platform roles, see [Using RBAC to define and apply permissions](https://docs.openshift.com/container-platform/4.15/authentication/using-rbac.html).
-* You have created a project in OpenShift Container Platform. For more information about creating a project in OpenShift Container Platform, see [Red Hat OpenShift Container Platform documentation](https://docs.openshift.com/container-platform/4.15/applications/projects/working-with-projects.html#odc-creating-projects-using-developer-perspective_projects).
-* You have installed the link:https://helm.sh/docs/intro/install[Helm CLI].
+* You are logged in to your {ocp-short} account.
+* A user with the {ocp-short} admin role has configured the appropriate roles and permissions within your project to create an application. For more information about {ocp-short} roles, see link:https://docs.openshift.com/container-platform/4.15/authentication/using-rbac.html[Using RBAC to define and apply permissions].
+* You have created a project in {ocp-short}. For more information about creating a project in {ocp-short}, see link:https://docs.openshift.com/container-platform/4.15/applications/projects/working-with-projects.html#odc-creating-projects-using-developer-perspective_projects[{ocp-brand-name} documentation].
+* You have installed the Helm CLI tool.
 
 .Procedure
 . Create and activate the _<rhdh>_ {ocp-short} project:
@@ -46,3 +46,5 @@ echo "https://redhat-developer-hub-$NAMESPACE.$CLUSTER_ROUTER_BASE"
 .Verification
 * Open the running {product-short} instance URL in your browser to use {product-short}.
 
+.Additional resources
+* For more information, see link:https://docs.openshift.com/container-platform/4.16/applications/working_with_helm_charts/installing-helm.html[Installing Helm] in the {ocp-short} documentation.


modules/installation/proc-installing-rhdh-on-openshift-with-helm-cli.adoc
```diff
@@ -5,10 +5,10 @@ You can use the Helm CLI to install {product} on {ocp-brand-name}.
...
-* You have installed the link:https://helm.sh/docs/intro/install[Helm CLI].
+* You are logged in to your {ocp-short} account.
+* A user with the {ocp-short} admin role has configured the appropriate roles and permissions within your project to create an application. For more information about {ocp-short} roles, see link:https://docs.openshift.com/container-platform/4.15/authentication/using-rbac.html[Using RBAC to define and apply permissions].
+* You have created a project in OpenShift Container Platform. For more information about creating a project in OpenShift Container Platform, see link:https://docs.openshift.com/container-platform/4.15/applications/projects/working-with-projects.html#odc-creating-projects-using-developer-perspective_projects[{ocp-brand-name} documentation].
+* You have installed the Helm CLI tool.
 
 .Procedure
 . Create and activate the _<rhdh>_ {ocp-short} project:
```
linfraze, line 14: why is RHDH formatted this way?

modules/installation/proc-installing-rhdh-on-openshift-with-helm-cli.adoc
```diff
@@ -5,10 +5,10 @@ You can use the Helm CLI to install {product} on {ocp-brand-name}.
...
-* You have created a project in OpenShift Container Platform. For more information about creating a project in OpenShift Container Platform, see [Red Hat OpenShift Container Platform documentation](https://docs.openshift.com/container-platform/4.15/applications/projects/working-with-projects.html#odc-creating-projects-using-developer-perspective_projects).
-* You have installed the link:https://helm.sh/docs/intro/install[Helm CLI].
+* You are logged in to your {ocp-short} account.
+* A user with the {ocp-short} admin role has configured the appropriate roles and permissions within your project to create an application. For more information about {ocp-short} roles, see link:https://docs.openshift.com/container-platform/4.15/authentication/using-rbac.html[Using RBAC to define and apply permissions].
+* You have created a project in {ocp-short}. For more information about creating a project in {ocp-short}, see link:https://docs.openshift.com/container-platform/4.15/applications/projects/working-with-projects.html#odc-creating-projects-using-developer-perspective_projects[{ocp-brand-name} documentation].
```
linfraze, line 10: a specific project?

modules/installation/proc-installing-rhdh-on-openshift-with-helm-cli.adoc
Preeticp: AFAIU, it is the name of the Namespace mentioned in the code block below that is the reason for the Italics.

linfraze: The italics are fine, but wondering about the brackets and whether this statement and that project name clearly communicates to the user what they are creating and activating here.

linfraze: Also, it is a Prerequisite that a project is already created in OCP - is step 1 a redundancy or are we asking them to create a new/different project here?

linfraze: @nickboldt I am going to merge this PR to go ahead and fix the original bugs / broken links, but would like to circle back to these other Qs and potentially create a separate PR if add'l changes are needed to improve clarity.

nickboldt: yes, you need a project / namespace, before you can install into the project / namespace.

Migrating these "set of manual instructions" to a "script that does it for you" will make it easier for the following people to validate the doc:
* QE
* SME
* doc writer
* end user
* customer

I realize there's a time commitment to running through the steps manually to see if they make sense, but would it make sense to have doc writers do more than just WRITE? for example if you log into the Dev Sandbox, and actually try out installing and deploying RHDH? Then it would be obvious that you need a namespace before you can deploy onto the cluster. :D

modules/installation/proc-installing-rhdh-on-openshift-with-helm-cli.adoc
nickboldt: any namespace or project you'd like, though there are recommendations for a standard name like "rhdh" or "developer-hub"

modules/installation/proc-installing-rhdh-on-openshift-with-helm-cli.adoc
```diff
@@ -46,3 +46,5 @@ echo "https://redhat-developer-hub-$NAMESPACE.$CLUSTER_ROUTER_BASE"
 .Verification
 * Open the running {product-short} instance URL in your browser to use {product-short}.
 
+.Additional resources
+* For more information, see link:https://docs.openshift.com/container-platform/4.16/applications/working_with_helm_charts/installing-helm.html[Installing Helm] in the {ocp-short} documentation.
```
nickboldt, line 50: curious why we didnt' just link to this from line 11:

```
* You have installed the link:https://docs.openshift.com/container-platform/4.16/applications/working_with_helm_charts/installing-helm.html[Helm tool].
```

modules/installation/proc-installing-rhdh-on-openshift-with-helm-cli.adoc
linfraze: @nickboldt Yeah, we need a project / namespace, but my question about redundancy. We have a Prereq that says that you must have a project and then step 1 is "Create a project" - are we talking about 2 separate projects or is Step 1 duplicating the Prereq?

linfraze: @nickboldt are you suggesting internal or customer use of such a script? Replacing manual steps with a script is a beautiful thought, but there are many considerations and it might not necessarily be the best solution, especially for customer use. For instance, does it require the user to clone a GitHub repo, especially one that contains non-RH-supported content? Who will maintain this script? How do we know that customers even want to use the script?

modules/installation/proc-installing-rhdh-on-openshift-with-helm-cli.adoc
linfraze: We might want to be more specific in the future and say something like _You have created a Red Hat Developer Hub project..._.

modules/installation/proc-installing-rhdh-on-openshift-with-helm-cli.adoc
linfraze: @nickboldt That is an option; I just followed the instructions in the Jira to add this as an Additional resource. I know there are some opinions in CCS that we should include all links in an Additional resources section and not in-line, but we don't follow that consistently in RHDH. 

@Preeticp is there a specific reason why you recommended adding this link in the Additional resources sections rather than in the Prerequisites section?

modules/installation/proc-installing-rhdh-on-openshift-with-helm-cli.adoc
nickboldt: the prereq is to have a namespace into which you can deploy the app.
the command in step 1 will either CREATE a new, or SELECT the existing namespace
```
oc new-project ${NAMESPACE} || oc project ${NAMESPACE}
```

So yes, the prereq is sort of redundant to the command. If it was me, I'd remove the prereq.

nickboldt: The script could be accessible within the docs as an artifact file or code block, or could live in a GH repo and be referenced by a raw URL link. 

In one case you'd click a link from the docs, download the script, review it, then run it.

In the other case you'd read the script in the docs, and then copy-paste-keywhack to run it. 

We already have the steps documented below this line (starting on line 19 and continuing to line 43. The Style Guide suggests you want to break each and every command into a separate copy-and-paste code block... which is ok, I guess, but not great UX for devs who want to just make things work quickly.

modules/installation/proc-installing-rhdh-on-openshift-with-helm-cli.adoc
Preeticp: Sorry for the late response, but we could link it to the prerequisite on line 11.

nickboldt: could you open a new PR @Preeticp  ? this was merged 3 weeks ago.

PR #388, opened 2024-07-24
Fix janus-idp-backstage-plugin-rbac path

sbem44:
The path for the community "janus-idp-backstage-plugin-rbac" is incorrect. It should not end with "-dynamic".


PatAKnight (approved):
Matches [showcase](https://github.com/janus-idp/backstage-showcase/blob/main/dynamic-plugins.default.yaml#L540), lgtm

PR #467, opened 2024-09-05
RHIDP-3440 Update product version to fix broken Pantheon builds

Gerry-Forde:
We are using `main` to preview 1.3 docs in Pantheon, so we need to update the product versions to match the Pantheon build version.

**Version(s):**
1.3

**Issue:**

**Link to docs preview:**

**Reviews:**
- [ ] Docs review: @ mention assignee

**Additional information:**

### DIFF
--- a/artifacts/attributes.adoc
+++ b/artifacts/attributes.adoc
@@ -10,9 +10,9 @@
 :product: Red Hat Developer Hub
 :product-short: Developer Hub
 :product-very-short: RHDH
-:product-version: 1.4
-:product-bundle-version: 1.4.0
-:product-chart-version: 1.4.0
+:product-version: 1.3
+:product-bundle-version: 1.3.0
+:product-chart-version: 1.3.0
 :product-backstage-version: 1.27.7
 :rhdeveloper-name: Red Hat Developer
 :rhel: Red Hat Enterprise Linux


hmanwani-rh:
@Gerry-Forde I was under the impression that we would use the stage for the 1.3 docs and the preview for the main docs. Did you intend to change the version in the 1.3 branch? Just to clarify, this PR is currently created against the main branch, where I believe it is appropriate to keep the version value as 1.4. what do you think?

Gerry-Forde:
@hmanwani-rh 
Thanks Heena, do we not want the Red Hat Developer Hub **1.3** _Preview_ in Pantheon to point to the `main` branch?

Maybe I misunderstood and we're not going to use the _Preview_ at all on this page?

There's a build error on this page for _Preview_ if Pantheon is pointing to the `main` branch (with the 1.4 product version) for _Preview_. 
Screenshot 2024-09-05 at 13 09 52

themr0c (approved):

nickboldt:
This is technically wrong since the 1.3.x branch is for 1.3 docs and the main branch will be for 1.4 docs, but I understand your process is to NOT branch for another few weeks.

Meanwhile my process is to automate the branch creation at the time of feature freeze, which is why the 1.3.x branch exists today, rather than in a few weeks. 

We could omit this repo from the list of repos to branch https://gitlab.cee.redhat.com/rhidp/rhdh/-/blob/rhdh-1-rhel-9/build/scripts/tagRelease.sh?ref_type=heads#L694 but then you'd have to create your branches manually.

Anyway, just wanted to give context for why the bump to 1.4 was done in main branch, and why the 1.3.x branch has version 1.3 as the doc/operator/helm chart version.

PR #623, opened 2024-10-15
chore: removed unused Gitlab authentication procedure

themr0c:
chore: removed unused and unsupported authentication procedure for Gittlab

**Version(s):** 1.3.1

**Issue:**

**Link to docs preview:**

**Reviews:**
- [ ] SME: @ mention assignee
- [ ] QE: @ mention assignee
- [ ] Docs review: @ mention assignee
- [ ] Additional review: @mention assignee (by writer)

**Additional information:**


hmanwani-rh:
@themr0c We may need this section in the future when GitLab is supported. I suggest we keep the module but exclude it for now.

themr0c:
The content has not been reworked along the authorization book revamp. We should not reintroduce it later in current state.

hmanwani-rh:
> The content has not been reworked along the authorization book revamp. We should not reintroduce it later in current state.


What if we need the existing information? We can always revamp it into the desired state as needed.

themr0c:
We can get the information back from history: this PR for example. Don't keep dead code in a project. It's confusing.

hmanwani-rh (approved):

PR #698, opened 2024-11-12
Added discover content

hmanwani-rh:
I’m creating this PR because the About RHDH title is already published on docs.redhat.com, but currently, it doesn’t contain any content. For now, I’m adding content to the page, which can be refined and improved later.

cc @jmagak


hmanwani-rh:
Closing this PR, since the title is now unpublished in Pantheon.

PR #734, opened 2024-11-26
RHIDP-3490-Updates the context

jmagak:

**Version(s):** 1.2

**Issue:** RHIDP-3490

### DIFF
--- a/modules/installation/proc-configuring-an-rhdh-instance-with-tls-in-kubernetes.adoc
+++ b/modules/installation/proc-configuring-an-rhdh-instance-with-tls-in-kubernetes.adoc
@@ -1,7 +1,7 @@
 [id="proc-configuring-an-rhdh-instance-with-tls-in-kubernetes_{context}"]
 = Configuring an {product-very-short} instance with a TLS connection in Kubernetes
 
-You can configure an {product-very-short} instance with a Transport Layer Security (TLS) connection in a Kubernetes cluster, such as an Azure Red Hat OpenShift (ARO) cluster, any cluster from a supported cloud provider, or your own cluster with proper configuration. However, You must use a public Certificate Authority (CA)-signed certificate to configure your Kubernetes cluster.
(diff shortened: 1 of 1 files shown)


modules/installation/proc-configuring-an-rhdh-instance-with-tls-in-kubernetes.adoc
```diff
@@ -1,7 +1,7 @@
 [id="proc-configuring-an-rhdh-instance-with-tls-in-kubernetes_{context}"]
 = Configuring an {product-very-short} instance with a TLS connection in Kubernetes
 
-You can configure an {product-very-short} instance with a Transport Layer Security (TLS) connection in a Kubernetes cluster, such as an Azure Red Hat OpenShift (ARO) cluster, any cluster from a supported cloud provider, or your own cluster with proper configuration. However, You must use a public Certificate Authority (CA)-signed certificate to configure your Kubernetes cluster.
+You can configure a {product-very-short} instance with a Transport Layer Security (TLS) connection in a Kubernetes cluster, such as an Azure Red Hat OpenShift (ARO) cluster, any cluster from a supported cloud provider, or your own cluster with proper configuration. Communication between the {product-very-short} and the cluster is enabled by using the Transport Layer Security (TLS). Users can encrypt data and verify identity. However, you must use a public Certificate Authority (CA)-signed certificate to configure your Kubernetes cluster.
```
shipsing, line 4: It is not correct. Communication does not happen between cluster and the RHDH instance. It is needed to secure the RHDH instance connection with other entities (could be third-party apps, external databases, or any other entity).
`Users can encrypt data and verify identity.` --- We should not make it this specific because there are other use cases as well. Maybe, we can clarify it with SMEs.

jmagak: Thank you, @shipsing. I am collecting input from the SME about it and also on the 'why' we configure RHDH instance with a TLS connection in a Kubernetes cluster.

PR #750, opened 2024-11-28
Include the support preface content only once in the entire doc set

themr0c:
* Change the content about support to a reference module and include in the About title.
* Remove the current preface about support from all titles.
* To avoid that the introduction paragraph has the *Preface* title, use the `article` document type, rather than `book`.

**Version(s):** 1.3.x, 1.4.x

**Issue:** https://issues.redhat.com/browse/RHIDP-5099


Preeticp:
@themr0c I like the reason why we are doing this, but can we get some reviews from either @Gerry-Forde or @hmanwani-rh as well before we merge this?

PR #788, opened 2024-12-11
Release-1.3 Release notes script fixes and enhancements

themr0c:
This PR adds fixes and enhancements to the release notes script on top of the release-1.3 pull request.

Enhancements
* Uses the attributes.adoc file to retrieve product version (y and z), rather than the YAML file that we forgot to update.
* Fixed the Jira search to be accurate on a z-stream. (The script was tested on a y-stream only, and had weird results on a z-stream).
* Uses only one content template, rather than 3.
* Generates only 6 files, rather than one file per Jira.
* In the Fixed issues section create one subtitle for each z-stream version.
* Double checked that the Jira searches are accurate.

**Version(s):** 1.3

**Issue:**


nickboldt:
Two comments on the redis fix...

1., why are we titling it "backport..." instead of "what the fix actually does" ?

2., the formatting with this update breaks the links from being bulleted lists 

Before:

image

After:

image

themr0c:
@nickboldt Your final vote?

nickboldt (approved):
with fixes for title of the redis bug and the formatting no longer mangled, +1

PR #789, opened 2024-12-12
main chore: release notes single-sourcing script fixes and enhancements

themr0c:
*    Uses the attributes.adoc file to retrieve product version (y and z), rather than the YAML file that we forgot to update.
*    Fixed the Jira search to be accurate on a z-stream. (The script was tested on a y-stream only, and had weird results on a z-stream).
*    Uses only one content template, rather than 3.
*    Generates only 6 files, rather than one file per Jira.
*    In the Fixed issues section create one subtitle for each z-stream version.
*     Double checked that the Jira searches are accurate.
* handle the case when the z-stream is 0.

**Version(s):** 1.5

**Issue:**


modules/release-notes/single-source-release-notes.py
```diff
@@ -13,109 +13,93 @@
...
+with open(attributes_file) as file:
+  attributes = {}
+  for lines in file:
+    if re.search(':', lines):
+      items = lines.rsplit(':', 2)
+      attributes[items[1]] = items[2].strip()
+
+product_version_minor = attributes["product-version"]
+product_version_patch = attributes["product-bundle-version"]
+
```
themr0c, line 52:
```suggestion
# Until 1.4 at least, Jira has no z-stream 0. Use the y-stream instead.
if re.search('.0$',product_version_patch):
  product_version_patch = product_version_minor

```

Fortune-Ndlovu (commented):
@themr0c Looks good, can you give me context as to why these enhancements are needed?

themr0c:
@Fortune-Ndlovu I updated the PR description to answer your question.

artifacts/attributes.adoc
```diff
@@ -10,9 +10,9 @@
 :product: Red Hat Developer Hub
 :product-short: Developer Hub
 :product-very-short: RHDH
-:product-version: 1.4
-:product-bundle-version: 1.4.0
-:product-chart-version: 1.4.0
+:product-version: 1.5
+:product-bundle-version: 1.5.0
+:product-chart-version: 1.5.0
 :product-backstage-version: 1.29.2
```
themr0c, line 16:
```suggestion
:product-backstage-version: 1.31.3
```

PR #868, opened 2025-01-21
Update the Helm procedure to use oc-mirror for air-gapped OCP environments

Fortune-Ndlovu:

**Version(s):**

This PR updates the air-gapped installation documentation for Helm using oc-mirror. The key updates include:

- Adding detailed steps for:
- Creating the ImageSetConfiguration file.
- Running the mirroring process.
- Transferring and extracting the archive in an air-gapped environment.
- Installing and verifying the Helm chart.

Curious about the visual output take a look at this gist of the update: https://gist.github.com/Fortune-Ndlovu/8b93a529ae6e79e7d79bf256d078bf38 

**Issue:**
Related Issue: https://issues.redhat.com/browse/RHIDP-5351 

Issue where the testing was verified: https://issues.redhat.com/browse/RHIDP-1723

**Preview** 
Deep link: https://redhat-developer.github.io/red-hat-developers-documentation-rhdh/pr-868/install-rhdh-air-gapped/#proc-install-rhdh-airgapped-environment-ocp-helm_title-install-rhdh-air-grapped


Fortune-Ndlovu:
Closing PR because docs member is already working on this PR update

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -6,7 +6,7 @@
 
 An air-gapped environment, also known as an air-gapped network or isolated network, ensures security by physically segregating the system or network. This isolation is established to prevent unauthorized access, data transfer, or communication between the air-gapped system and external sources.
 
-You can install {product} in an air-gapped environment to ensure security and meet specific regulatory requirements.
+This documentation outlines the steps to perform an air-gapped installation of Helm charts using the `oc-mirror` tool. The process ensures that all necessary resources are mirrored and deployed within a restricted environment.
```
linfraze, line 9: Remove self-referential language (_this documentation outline the steps_...).
We strive to write customer-centric doc rather than doc- or product-centric doc. \
* Customer-centric: _you can..._
* Doc-centric: _this documentation..._
* Product-centric: _this feature allows users to..._

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -17,7 +17,8 @@ To install {product-short} in an air-gapped environment, you must have access to
 * You have access to the {ocp-brand-name} image registry of your cluster. For more information about exposing the image registry, see the {ocp-brand-name} documentation about https://docs.openshift.com/container-platform/{ocp-version}/registry/securing-exposing-registry.html[Exposing the registry].
 * You have installed the {openshift-cli} on your workstation.
 * You have installed the `podman` command line tools on your workstation.
-* You you have an account in https://developers.redhat.com/[{rhdeveloper-name}] portal.
+* You have installed the `oc-mirror`, review https://docs.openshift.com/container-platform/4.17/disconnected/mirroring/installing-mirroring-disconnected.html#installation-oc-mirror-installing-plugin_installing-mirroring-disconnected[Installing the oc-mirror OpenShift CLI plugin].
```
linfraze, line 20: For more information, see [link[

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -59,80 +60,62 @@ podman login registry.redhat.io
...
-podman pull registry.redhat.io/rhel9/postgresql-15:latest
-----
-
-. Push both images to the internal {ocp-short} image registry by running the following commands:
-+
-[source,terminal,source,subs="attributes+"]
-----
-podman push --remove-signatures registry.redhat.io/rhdh/rhdh-hub-rhel9:{product-version} default-route-openshift-image-registry.<hostname>/<project_name>/rhdh-hub-rhel9:{product-version}
-----
+Example *imageset-config-ref.yaml*
```
linfraze, line 66: . Create an `ImageSetConfiguration` file to specify the resources that you want to mirror. For example:

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -59,80 +60,62 @@ podman login registry.redhat.io
...
+          - name: redhat-developer-hub
+            version: "1.4.1"
+  additionalImages:
+    - name: registry.redhat.io/rhdh/rhdh-hub-rhel9:1.4-1737055846
+    - name: registry.redhat.io/rhel9/postgresql-15:latest
+----
+. Run the mirroring process on the ImageSetConfiguration file. Use the
+`oc-mirror` command to mirror the specified resources based on the
+configuration file. his command generates a `.tar` archive that contains
+all the mirrored resources.
```
linfraze, line 90: Need to better understand this process so we can write it more clearly

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -59,80 +60,62 @@ podman login registry.redhat.io
...
 [source,terminal]
 ----
-oc get imagestream -n <project_name>
+oc-mirror --config=imageset-config-ref.yaml **** file://mirror-archive
 ----
-
-. Enable local image lookup for both images by running the following commands:
+. Transfer the generated archive (e.g.,
+`mirror++_++seq1++_++000000.tar`) to the air-gapped environment using
+`scp` or another file transfer tool.
```
linfraze, line 98: Remove line breaks

Add _For example:_

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -59,80 +60,62 @@ podman login registry.redhat.io
...
 +
 [source,terminal]
 ----
-oc set image-lookup postgresql-15
+scp ./mirror-archive/mirror_seq1_000000.tar /tmp
 ----
+. Extract the contents of the archive into a directory before proceeding
+with deployment.
```
linfraze, line 105: Remove line breaks 

How is the content extracted? What actions does the user take to do this?

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -59,80 +60,62 @@ podman login registry.redhat.io
...
 +
 [source,terminal]
 ----
-oc set image-lookup  rhdh-hub-rhel9
+tar -xvf /mirror-archive/mirror_seq1_000000.tar -C /tmp
 ----
-
-. Go to *YAML view* and update the `image` section for `backstage` and `postgresql` using the following values:
+. Deploy the Helm chart in the air-gapped environment using the `helm`
```
linfraze, line 111: Where is the command that the user can copy and paste?

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -59,80 +60,62 @@ podman login registry.redhat.io
...
-      repository: postgresql-15
-      tag: latest
+helm install rhdh ./mirror-archive/oc-mirror-workspace/src/charts/redhat-developer-hub-1.4.1.tgz --namespace rhdh --create-namespace
 ----
---
 
-. Install the {product} using Helm chart.
+Verify that all resources are successfully deployed and running.
+Depending on the cluster’s router base, ensure the route matches the
+expected URL. If necessary, update the route.
```
linfraze, line 121: Convert this to a Verification section per CCS style guidelines

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
Fortune-Ndlovu: The `ImageSetConfiguration` file specifies what resources to mirror. In the example provided, it includes Helm charts (e.g., redhat-developer-hub version 1.4.1 from the openshift-charts repository).

Initiating the `oc-mirror` command to processes the configuration file. This command retrieves the specified resources and packages them into a `.tar` archive.

The `.tar` archive is then used to transfer the mirrored resources to an air-gapped environment.

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
Fortune-Ndlovu: There are no unexpected line breaks in the text.  Its just how AsciiDoc's format code blocks.

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
Fortune-Ndlovu: There are no unexpected line breaks in the text.  It's just how AsciiDoc's format code blocks.

Good Question - Actions the user takes to do this:

The content of the archive is extracted using the `tar` command after transferring it to the air-gapped environment. Once the archive (e.g., mirror_seq1_000000.tar) is successfully transferred, the user runs the following command to extract its contents into a specified directory:

- tar -xvf /mirror-archive/mirror_seq1_000000.tar -C /tmp

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
Fortune-Ndlovu: Where is the command that the user can copy and paste?
[source,terminal]
----
scp ./mirror-archive/mirror_seq1_000000.tar /tmp
----

What is the code sample showing and what does the user do with it?
The code sample demonstrates the process of transferring the generated archive (mirror_seq1_000000.tar) to the air-gapped environment using scp. Once transferred, the user extracts the contents of the archive using the tar command. This step is necessary to make the mirrored resources (such as container images, Helm charts, or Operator catalogs) accessible in the air-gapped environment.

Finally, the user uses these extracted resources to deploy and configure their applications or clusters in the disconnected environment.

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -59,80 +60,62 @@ podman login registry.redhat.io
...
-[source,yaml]
+[source,terminal]
 ----
-upstream:
-  postgresql:
-    image:
-      registry: ""
-      repository: postgresql-15
-      tag: latest
+helm install rhdh ./mirror-archive/oc-mirror-workspace/src/charts/redhat-developer-hub-1.4.1.tgz --namespace rhdh --create-namespace
```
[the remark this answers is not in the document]
Fortune-Ndlovu: We mention the `helm` installation command lastly because at this moment the user is able and should deploy the Helm chart in the air-gapped environment.

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
Fortune-Ndlovu: Thanks for the clarity @linfraze I have reverted to the original statement.

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -59,80 +60,59 @@ podman login registry.redhat.io
...
-====
-If an x509 error occurs, verify that you have link:https://access.redhat.com/solutions/6088891[installed the CA certificate used for {ocp-short} routes on your system].
-====
-
-. Use the following command to verify that both images are present in the internal {ocp-short} registry:
+apiVersion: mirror.openshift.io/v1alpha2
+kind: ImageSetConfiguration
+storageConfig:
+  local:
+    path: ./mirror-output
```
rm3l, line 73: IMO, we should distinguish between mirroring to a mirror registry in a partially disconnected environment or in a fully disconnected environment.
In a partially disconnected disconnected environment, users can mirror the image set directly to the target mirror registry. So they won't need to mirror everything to disk first (using this `storageConfig.local.path` field). Because it might result in downloading several hundreds of gigabytes of data to disk, which is useless in that case IMO.
What you are describing here makes sense in a fully disconnected environment, where users would first mirror the image set to disk, then from the disk to the target registry.

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -59,80 +60,59 @@ podman login registry.redhat.io
...
+mirror:
+  helm:
+    repositories:
+      - name: openshift-charts
+        url: https://charts.openshift.io
+        charts:
+          - name: redhat-developer-hub
+            version: "1.4.1"
```
rm3l, line 81: This should be the product full version variable, not hardcoded to 1.4.1.

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
rm3l: IMO, for a fully disconnected env, the next steps once the image set has been mirrored to disk should be something like "Transfer the archive to the disconnected environment".
They might use `scp`, `rsync` or even physically move from a flash drive to their disconnected environment. This is up to them, and we cannot know that.

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
linfraze: Hm, it looks regular text rather than a code block and should be on one line in the adoc. We want to maintain the integrity of the files by eliminating any unnecessary line breaks, but the most important thing is that everything renders correctly in the end product.

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -17,7 +17,8 @@ To install {product-short} in an air-gapped environment, you must have access to
 * You have access to the {ocp-brand-name} image registry of your cluster. For more information about exposing the image registry, see the {ocp-brand-name} documentation about https://docs.openshift.com/container-platform/{ocp-version}/registry/securing-exposing-registry.html[Exposing the registry].
 * You have installed the {openshift-cli} on your workstation.
 * You have installed the `podman` command line tools on your workstation.
-* You you have an account in https://developers.redhat.com/[{rhdeveloper-name}] portal.
+* You have installed the `oc-mirror`, for more information see https://docs.openshift.com/container-platform/4.17/disconnected/mirroring/installing-mirroring-disconnected.html#installation-oc-mirror-installing-plugin_installing-mirroring-disconnected[Installing the oc-mirror OpenShift CLI plugin].
```
linfraze, line 20:
```suggestion
* You have installed the oc-mirror {openshift-cli} plugin, for more information see https://docs.openshift.com/container-platform/4.17/disconnected/mirroring/installing-mirroring-disconnected.html#installation-oc-mirror-installing-plugin_installing-mirroring-disconnected[Installing the oc-mirror OpenShift CLI plugin].
```

linfraze: Personally, I think it should be formatted as `oc-mirror` but the OCP docs don't monospace it, so we'll do what they do

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -59,80 +60,59 @@ podman login registry.redhat.io
...
+mirror:
+  helm:
+    repositories:
+      - name: openshift-charts
+        url: https://charts.openshift.io
+        charts:
+          - name: redhat-developer-hub
+            version: "1.4.1"
+----
+. Initiate the mirroring process using the `ImageSetConfiguration` file. The `oc-mirror` command mirrors the specified resources as defined in the configuration file and generates a `.tar` archive containing all the mirrored resources.
```
linfraze, line 83: `. Initiate the mirroring process using the `ImageSetConfiguration` file. `

I'll need to understand this better so we can communicate this in a way that is more clear, specific, and in alignment with certain doc standards.

linfraze, line 83:
```suggestion
. Mirror the resources specified in the`ImageSetConfiguration` file by running the following command.
```
Make the following a Note or move to the conceptual portion of the module before the procedure: 

`The `oc-mirror` command mirrors the specified resources as defined in the configuration file and generates a `.tar` archive containing all the mirrored resources.`

linfraze: Is the changed step still technically accurate?

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -59,80 +60,59 @@ podman login registry.redhat.io
...
 [source,terminal]
 ----
-oc get imagestream -n <project_name>
+oc-mirror --config=mirror-config.yaml file://mirror-archive
 ----
-
-. Enable local image lookup for both images by running the following commands:
+. Transfer the generated archive (e.g.,
+`mirror++_++seq1++_++000000.tar`) to the air-gapped environment using
+`scp` or another file transfer tool.
```
linfraze, line 91:
```suggestion
. Transfer the generated `.tar` archive file (for example, `mirror++_++seq1++_++000000.tar`) to the air-gapped environment using `scp` or another file transfer tool. For example:
```

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -59,80 +60,59 @@ podman login registry.redhat.io
...
 +
 [source,terminal]
 ----
-oc set image-lookup postgresql-15
+scp ./mirror-archive/mirror_seq1_000000.tar /tmp
 ----
+. Extract the contents of the archive into a directory before proceeding
+with deployment.
```
linfraze, line 98:
```suggestion
. Extract the contents of the archive into a directory by running the following command:
```

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -59,80 +60,59 @@ podman login registry.redhat.io
...
 +
 [source,terminal]
 ----
-oc set image-lookup  rhdh-hub-rhel9
+tar -xvf /mirror-archive/mirror_seq1_000000.tar -C /tmp
 ----
-
-. Go to *YAML view* and update the `image` section for `backstage` and `postgresql` using the following values:
+. Deploy the Helm chart in the air-gapped environment using the `helm`
+command.
```
linfraze, line 105:
```suggestion
. Deploy the Helm chart in the air-gapped environment by running the following command:
```

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -59,80 +60,59 @@ podman login registry.redhat.io
...
-      repository: postgresql-15
-      tag: latest
-----
---
+.Verification
+1. Verify that all resources are successfully deployed and running:
+   - Check the status of your deployments and pods in the cluster.
+2. Validate the route:
+   - Ensure the route matches the expected URL based on the cluster's router base.
+   - If necessary, update the route to align with your expected configuration.
```
linfraze, line 117:
```suggestion
. To verify that all resources are successfully deployed and running, check the status of your deployments and pods in the cluster. 
. To validate the route, ensure that the route matches the expected URL based on the router base of the cluster. 
.. If necessary, update the route to align with your expected configuration.
```

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
linfraze: The conceptual information in the module is very generic and replicates the information in the higher-level module. Replace it with more specific information - if we split into two modules (one for fully disconnected, one for partially disconnected), it would be a good idea to give specific conceptual information for both of those scenarios and explain how / why the process is unique to each use case. 

Note: we've received general feedback that customers want more conceptual info (to better understand what and why) in Red Hat docs.

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
linfraze: How the procedures are different: 

**Fully disconnected:** 
1. mirror everything to disk using this storageConfig.local.path field
2. transfer the archive to the disconnected environment
3. mirror from the disk to the target registry

**Partially disconnected:**
1. mirror the image set directly to the target mirror registry. 
(Skip steps 1 & 2 because it might result in unnecessarily downloading several hundreds of gigabytes of data to disk)

linfraze: Proper CCS modularization standards would have us create two separate modules, but I'm open including both procedures in the same module for now and modularizing under a separate PR in the future. 

So we would have the following structure: 

**= Installing {product} in an air-gapped environment with the Helm Chart**

[Add conceptual information specific to installing RHDH in a air-gapped environment using the Helm chart. Inform the user that they can do this whether they are in a fully or partially disconnected environment and that the process is slightly different depending on their use case] - Lindsey will draft this using info from the SME. 

**== Installing {product} in a partially disconnected environment with the Helm Chart**

[conceptual and procedural info for this use case]

**== Installing {product} in a fully disconnected environment with the Helm Chart**

[conceptual and procedural info for this use case]

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
linfraze: {product-version} renders the RHDH version that the customer is looking at docs for

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
linfraze: See [this related comment ](https://github.com/redhat-developer/red-hat-developers-documentation-rhdh/pull/868#discussion_r1924904601)also

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
Fortune-Ndlovu: Agreed!

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
linfraze: nvm, don't want to scope creep

reminder to update conceptual info with a future PR

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
linfraze: Are the prerequisites different? Should we mention the need for some connectivity to complete the partially connected procedure?

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -57,82 +57,142 @@ podman login -u kubeadmin -p $(oc whoami -t) $REGISTRY_HOST
...
+          - name: redhat-developer-hub
+            version: "{product-version}"
+----
+. Mirror the resources specified in the`ImageSetConfiguration` file directly to the target registry by running the following command:
 +
-[source,terminal,source,subs="attributes+"]
+[source,terminal]
 ----
-podman pull registry.redhat.io/rhdh/rhdh-hub-rhel9:{product-version}
+oc-mirror --config=mirror-config.yaml file://mirror-archive
```
rm3l, line 82: @Fortune-Ndlovu Do you confirm this works as expected by mirroring the right images? I just tested it, and it seems it is not able to detect the images:
```
$ oc mirror -config=.asoro/airgap/my-image-set-configuration.yaml file://mirror-archive
Checking push permissions for registry.localhost:5000
Found: oc-mirror-workspace/src/publish
Found: oc-mirror-workspace/src/v2
Found: oc-mirror-workspace/src/charts
Found: oc-mirror-workspace/src/release-signatures
backend is not configured in .asoro/airgap/my-image-set-configuration.yaml, using stateless mode
backend is not configured in .asoro/airgap/my-image-set-configuration.yaml, using stateless mode
No metadata detected, creating new workspace
Pulling chart redhat-developer-hub
error image list []
No new images detected, process stopping
```

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -57,82 +57,142 @@ podman login -u kubeadmin -p $(oc whoami -t) $REGISTRY_HOST
...
+  local:
+    path: ./mirror-output
+mirror:
+  helm:
+    repositories:
+      - name: openshift-charts
+        url: https://charts.openshift.io
+        charts:
+          - name: redhat-developer-hub
+            version: "{product-version}"
```
rm3l, line 76: Not sure if it is the right variable or an issue with the preview, but it seems this is not getting replaced in the final preview:
image

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -57,82 +57,142 @@ podman login -u kubeadmin -p $(oc whoami -t) $REGISTRY_HOST
...
 ----
-oc set image-lookup postgresql-15
+oc-mirror --config=mirror-config.yaml file://mirror-archive
 ----
+. Transfer the generated `.tar` archive file (for example, `mirror++_++seq1++_++000000.tar`) to the air-gapped environment using `scp` or another file transfer tool. For example:
 +
 [source,terminal]
 ----
-oc set image-lookup  rhdh-hub-rhel9
+scp ./mirror-archive/mirror_seq1_000000.tar /tmp
```
rm3l, line 176: This example is a bit misleading IMO. It should emphasize transferring to a remote host IMO.
I think we should remove this example. The bullet point could simply be "Transfer the generated `.tar` archive file (for example, `mirror++_++seq1++_++000000.tar`) to the air-gapped environment".
It is hard to know exactly how people transfer bits to their fully disconnected envs.
WDYT?

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -57,82 +57,142 @@ podman login -u kubeadmin -p $(oc whoami -t) $REGISTRY_HOST
...
 ----
-
-. Go to *YAML view* and update the `image` section for `backstage` and `postgresql` using the following values:
+. Extract the contents of the archive into a directory (for example, `/tmp`) by running the `tar` command. For example:
```
rm3l, line 178: Maybe we should clarify that this extraction needs to be happen in the disconnected network. So the host to which the archive was transferred. Otherwise, it looks like this can be run locally.

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -57,82 +57,142 @@ podman login -u kubeadmin -p $(oc whoami -t) $REGISTRY_HOST
...
 ----
-podman pull registry.redhat.io/rhel9/postgresql-15:latest
+helm install rhdh ./mirror-archive/oc-mirror-workspace/src/charts/redhat-developer-hub-{product-version}.tgz --namespace rhdh --create-namespace
 ----
 
-. Push both images to the internal {ocp-short} image registry by running the following commands:
+.Verification
+. To verify that all resources are successfully deployed and running, check the status of your deployments and pods in the cluster.
+. To validate the route, ensure that the route matches the expected URL based on the router base of the cluster.
+.. If necessary, update the route to align with your expected configuration.
```
rm3l, line 94: How do we expect users to validate and update this?
How about using a similar procedure as what we have for "[Deploying Developer Hub on OCP with the Helm CLI](https://docs.redhat.com/en/documentation/red_hat_developer_hub/1.4/html/installing_red_hat_developer_hub_on_openshift_container_platform/assembly-install-rhdh-ocp-helm#proc-install-rhdh-ocp-helm-cli_assembly-install-rhdh-ocp-helm)", by providing the command for the user to get the right router base URL? Like so:

```sh
CLUSTER_ROUTER_BASE=$(oc get route console -n openshift-console -o=jsonpath='{.spec.host}' | sed 's/^[^.]*\.//')
helm install redhat-developer-hub./mirror-archive/oc-mirror-workspace/src/charts/redhat-developer-hub-1.4.1.tgz \
    --set global.clusterRouterBase="$CLUSTER_ROUTER_BASE"
```

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
linfraze: Thanks for pointing that out. We need to make sure we include `subs="attributes+"` in the [source] tags

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -57,82 +57,142 @@ podman login -u kubeadmin -p $(oc whoami -t) $REGISTRY_HOST
...
+      - name: openshift-charts
+        url: https://charts.openshift.io
+        charts:
+          - name: redhat-developer-hub
+            version: "{product-version}"
+----
+. Mirror the resources specified in the`ImageSetConfiguration` file directly to the target registry by running the following command:
 +
-[source,terminal,source,subs="attributes+"]
+[source,terminal]
```
linfraze, line 80:
```suggestion
[source,terminal,subs="attributes+"]
```

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -57,82 +57,142 @@ podman login -u kubeadmin -p $(oc whoami -t) $REGISTRY_HOST
 podman login registry.redhat.io
 ----
 +
-For more information about registry authentication, see https://access.redhat.com/RegistryAuthentication[{company-name} Container Registry Authentication].
-
-. Pull {product-short} and PostgreSQL images from https://catalog.redhat.com/software/containers/search[{company-name} Image registry] to your workstation, by running the following commands:
+. Create an `ImageSetConfiguration` file to specify the resources that you want to mirror. For example:
++
+[source,terminal]
```
linfraze, line 62:
```suggestion
[source,terminal,subs="attributes+"]
```

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -57,82 +57,142 @@ podman login -u kubeadmin -p $(oc whoami -t) $REGISTRY_HOST
...
+* You have installed the `podman` command line tools on your workstation.
+* You have installed the oc-mirror {openshift-cli} plugin, for more information see https://docs.openshift.com/container-platform/4.17/disconnected/mirroring/installing-mirroring-disconnected.html#installation-oc-mirror-installing-plugin_installing-mirroring-disconnected[Installing the oc-mirror OpenShift CLI plugin].
+* You have an account in https://developers.redhat.com/[{rhdeveloper-name}] portal.
+
+.Procedure
+
+. Log in to your {ocp-short} account using the {openshift-cli} by running the following command:
 +
-[source,terminal,source,subs="attributes+"]
+[source,terminal]
```
linfraze, line 114:
```suggestion
[source,terminal,subs="attributes+"]
```

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -57,82 +57,142 @@ podman login -u kubeadmin -p $(oc whoami -t) $REGISTRY_HOST
...
 ----
-oc get imagestream -n <project_name>
+podman login registry.redhat.io
 ----
-
-. Enable local image lookup for both images by running the following commands:
++
+. Create an `ImageSetConfiguration` file to specify the resources that you want to mirror. For example:
++
+[source,terminal]
```
linfraze, line 150:
```suggestion
[source,terminal,subs="attributes+"]
```

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -57,82 +57,142 @@ podman login -u kubeadmin -p $(oc whoami -t) $REGISTRY_HOST
...
+            version: "{product-version}"
+----
+. Mirror the resources specified in the`ImageSetConfiguration` file by running the following command:
 +
 [source,terminal]
 ----
-oc set image-lookup postgresql-15
+oc-mirror --config=mirror-config.yaml file://mirror-archive
 ----
+. Transfer the generated `.tar` archive file (for example, `mirror++_++seq1++_++000000.tar`) to the air-gapped environment using `scp` or another file transfer tool. For example:
```
linfraze, line 172:
```suggestion
. Transfer the generated `.tar` archive file (for example, `mirror++_++seq1++_++000000.tar`) to the air-gapped environment.
```

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
linfraze, line 178:
```suggestion
. In your air-gapped environment, extract the contents of the archive into a directory (for example, `/tmp`) by running the `tar` command. For example:
```

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -57,82 +57,142 @@ podman login -u kubeadmin -p $(oc whoami -t) $REGISTRY_HOST
...
-. Pull {product-short} and PostgreSQL images from https://catalog.redhat.com/software/containers/search[{company-name} Image registry] to your workstation, by running the following commands:
+. Create an `ImageSetConfiguration` file to specify the resources that you want to mirror. For example:
++
+[source,terminal]
+----
+apiVersion: mirror.openshift.io/v1alpha2
+kind: ImageSetConfiguration
+storageConfig:
+  local:
+    path: ./mirror-output
```
linfraze, line 68: > Replace path: ./mirror-output with the actual path on your system

use variables and/or definitions to communicate this to the user ([like this](https://github.com/redhat-developer/red-hat-developers-documentation-rhdh/blob/42934116d39969bb1d65d5e101c66abc92dd2890/modules/observe/proc-admin-enabling-metrics-ocp-operator.adoc#L44))

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
rm3l: Whoops - I just noticed this comment. Yep, sounds good now. Thanks.

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -17,18 +17,18 @@ To install {product-short} in an air-gapped environment, you must have access to
 * You have access to the {ocp-brand-name} image registry of your cluster. For more information about exposing the image registry, see the {ocp-brand-name} documentation about https://docs.redhat.com/en/documentation/openshift_container_platform/{ocp-version}/html-single/registry/index#securing-exposing-registry[Exposing the registry].
 * You have installed the {openshift-cli} on your workstation.
 * You have installed the `podman` command line tools on your workstation.
```
rm3l, line 19: I think podman is not strictly required (oc-mirror can actually retrieve registry credentials from various places - see https://github.com/openshift/oc-mirror/blob/main/docs/usage.md#authentication ).
How about adding as prerequisite that they should be logged into their target mirror registry and have permission to push images there. We can link to [`Configuring credentials that allow images to be mirrored`](https://docs.openshift.com/container-platform/4.17/disconnected/mirroring/installing-mirroring-disconnected.html#installation-adding-registry-pull-secret_installing-mirroring-disconnected) for further details.

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -17,18 +17,18 @@ To install {product-short} in an air-gapped environment, you must have access to
...
-
-. Log in to your {ocp-short} account using the {openshift-cli}, by running the following command:
+. Log in to your {ocp-short} account using the {openshift-cli} by running the following command:
 +
 [source,terminal]
 ----
 oc login -u <user> -p <password> https://api.<hostname>:6443
 ----
 
-. Log in to the {ocp-short} image registry using the `podman` command line tool, by running the following command:
```
rm3l, line 31: See my previous comment. I think we can remove these podman commands (against the OCP image registry and against `registry.redhat.io`), and maybe link to https://docs.openshift.com/container-platform/4.17/disconnected/mirroring/installing-mirroring-disconnected.html#installation-adding-registry-pull-secret_installing-mirroring-disconnected in the prerequisites?

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -57,82 +57,149 @@ podman login -u kubeadmin -p $(oc whoami -t) $REGISTRY_HOST
...
-For more information about registry authentication, see https://access.redhat.com/RegistryAuthentication[{company-name} Container Registry Authentication].
-
-. Pull {product-short} and PostgreSQL images from https://catalog.redhat.com/software/containers/search[{company-name} Image registry] to your workstation, by running the following commands:
+. Create an `ImageSetConfiguration` file to specify the resources that you want to mirror. For example:
++
+[source,terminal]
+----
+apiVersion: mirror.openshift.io/v1alpha2
+kind: ImageSetConfiguration
+storageConfig:
```
rm3l, line 66: @Fortune-Ndlovu I think this `storageConfig` section is not needed and can be bypassed if we are mirroring directly to the target mirror registry.
`oc-mirror` creates a folder named `oc-mirror-workspace` in the current directory.

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -57,82 +57,149 @@ podman login -u kubeadmin -p $(oc whoami -t) $REGISTRY_HOST
...
+          - name: redhat-developer-hub
+            version: "{product-version}"
+----
+. Mirror the resources specified in the`ImageSetConfiguration` file directly to the target registry by running the following command:
 +
-[source,terminal,source,subs="attributes+"]
+[source,terminal]
 ----
-podman pull registry.redhat.io/rhdh/rhdh-hub-rhel9:{product-version}
+oc-mirror --config=mirror-config.yaml file://mirror-archive
```
rm3l, line 82:
```suggestion
oc-mirror --config=mirror-config.yaml docker://registry.example:5000
```

We are mirroring directly to the target registry in a partially disconnected environment. Otherwise, the images will never be pushed to the target mirror registry.
(This command is just an example - `registry.example:5000` should be replaced by the user)

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
rm3l: I've tried it again but it didn't work with my version of oc-mirror. I've reported https://github.com/openshift/oc-mirror/issues/1037 and also contributed a PR to fix it: https://github.com/openshift/oc-mirror/pull/1038

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
rm3l: Also, as verification procedure, the `oc-mirror` command above should end up with something like this once done:
```
[...]
Writing image mapping to oc-mirror-workspace/results-1738070846/mapping.txt
Writing ICSP manifests to oc-mirror-workspace/results-1738070846
```

The user needs to keep track of the directory mentioned since this contains the Helm Chart to apply.

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -57,82 +57,149 @@ podman login -u kubeadmin -p $(oc whoami -t) $REGISTRY_HOST
...
 ----
+. In your air-gapped environment, deploy the Helm chart to the `rhdh` namespace by running the `helm install` command. For example:
```
rm3l, line 84:
```suggestion
. Locate the `ImageContentSourcePolicy` (ICSP) manifest created by the `oc-mirror` command above and apply it against the cluster. For example:
+
[source,terminal,subs="attributes+"]
----
oc create -f oc-mirror-workspace/results-1738070846/imageContentSourcePolicy.yaml
----

. In your air-gapped environment, deploy the Helm chart to your current namespace by running the `helm install` command. For example:
```

I think there is a step missing before trying to deploy the Helm Chart. The user should apply the `ImageContentSourcePolicy` resource created by the `oc-mirror` command.
Without this, image mirroring won't work at all on OCP (because `oc-mirror` did not change the image refs in the Helm Chart it downloaded).

Also, the Helm Chart can be deployed in any namespace.

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -57,82 +57,149 @@ podman login -u kubeadmin -p $(oc whoami -t) $REGISTRY_HOST
...
 +
-[source,terminal,source,subs="attributes+"]
+[source,terminal,subs="attributes+"]
 ----
-podman pull registry.redhat.io/rhel9/postgresql-15:latest
+helm install rhdh ./mirror-archive/oc-mirror-workspace/src/charts/redhat-developer-hub-{product-version}.tgz --namespace rhdh --create-namespace
```
rm3l, line 88:
```suggestion
CLUSTER_ROUTER_BASE=$(oc get route console -n openshift-console -o=jsonpath='{.spec.host}' | sed 's/^[^.]*\.//') helm install oc-mirror-workspace/results-1738070846/charts/redhat-developer-hub-{product-version}.tgz \
    --set global.clusterRouterBase="$CLUSTER_ROUTER_BASE"
```

With the command above, the verification below is not needed.

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -57,82 +57,149 @@ podman login -u kubeadmin -p $(oc whoami -t) $REGISTRY_HOST
...
+
+== Installing {product} in a fully disconnected environment with the Helm chart
+
+You can use the Helm chart to install {product} by mirroring specified resources and transferring them to your air-gapped environment without any connection to the internet.
+
+.Prerequisites
+
+* You have installed {ocp-brand-name} {ocp-version-min} or later.
+* You have access to the `registry.redhat.io`.
+* You have access to the {ocp-brand-name} image registry for your cluster. For more information about exposing the image registry, see the {ocp-brand-name} documentation about https://docs.openshift.com/container-platform/{ocp-version}/registry/securing-exposing-registry.html[Exposing the registry].
```
rm3l, line 104: Same comment: https://github.com/redhat-developer/red-hat-developers-documentation-rhdh/pull/868#discussion_r1932122427

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -57,82 +57,149 @@ podman login -u kubeadmin -p $(oc whoami -t) $REGISTRY_HOST
...
+* You have installed the {openshift-cli} on your workstation.
+* You have installed the `podman` command line tools on your workstation.
```
rm3l, line 106: Same comment about Podman: https://github.com/redhat-developer/red-hat-developers-documentation-rhdh/pull/868#discussion_r1932132586

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -57,82 +57,149 @@ podman login -u kubeadmin -p $(oc whoami -t) $REGISTRY_HOST
...
+* You have installed the oc-mirror {openshift-cli} plugin, for more information see https://docs.openshift.com/container-platform/4.17/disconnected/mirroring/installing-mirroring-disconnected.html#installation-oc-mirror-installing-plugin_installing-mirroring-disconnected[Installing the oc-mirror OpenShift CLI plugin].
+* You have an account in https://developers.redhat.com/[{rhdeveloper-name}] portal.
+
+.Procedure
+
+. Log in to your {ocp-short} account using the {openshift-cli} by running the following command:
```
rm3l, line 112: In a fully disconnected setup, the user does not need to be connected to the cluster yet. They will be mirroring everything to disk first from the workstation. They will be required to log in to the target cluster and mirror registry from the disconnected network.

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -57,82 +57,149 @@ podman login -u kubeadmin -p $(oc whoami -t) $REGISTRY_HOST
...
 +
-[source,terminal,source,subs="attributes+"]
+[source,terminal]
 ----
-podman push --remove-signatures registry.redhat.io/rhdh/rhdh-hub-rhel9:{product-version} default-route-openshift-image-registry.<hostname>/<project_name>/rhdh-hub-rhel9:{product-version}
+oc login -u <user> -p <password> https://api.<hostname>:6443
 ----
+
+. Log in to the {ocp-short} image registry using the `podman` command line tool by running the following command:
```
rm3l, line 119: Same point related to the OCP image registry: https://github.com/redhat-developer/red-hat-developers-documentation-rhdh/pull/868#discussion_r1932135022

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -57,82 +57,149 @@ podman login -u kubeadmin -p $(oc whoami -t) $REGISTRY_HOST
...
+        charts:
+          - name: redhat-developer-hub
+            version: "{product-version}"
+----
+. Mirror the resources specified in the`ImageSetConfiguration` file by running the following command:
 +
 [source,terminal]
 ----
-oc set image-lookup postgresql-15
+oc-mirror --config=mirror-config.yaml file://mirror-archive
```
rm3l, line 170: As verification, we can maybe say that the command should end up with something like:
```
[...]
Creating archive /path/to/mirror-archive/mirror_seq1_000000.tar
```

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -57,82 +57,149 @@ podman login -u kubeadmin -p $(oc whoami -t) $REGISTRY_HOST
...
 ----
+. Transfer the generated `.tar` archive file (for example, `mirror++_++seq1++_++000000.tar`) to the air-gapped environment using `scp` or another file transfer tool. For example:
 +
 [source,terminal]
 ----
-oc set image-lookup  rhdh-hub-rhel9
+scp ./mirror-archive/mirror_seq1_000000.tar /tmp
```
rm3l, line 176:
```suggestion
scp ./mirror-archive/mirror_seq1_000000.tar [user@]host:/tmp/mirror_seq1_000000.tar
```

If we want to keep this example, I think it should highlight that this is a remote transfer.

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
rm3l: @Fortune-Ndlovu According to https://docs.openshift.com/container-platform/4.17/disconnected/mirroring/installing-mirroring-disconnected.html#oc-mirror-disk-to-mirror_installing-mirroring-disconnected , it is not needed to extract the contents. I've just tested it, and `oc-mirror` can read the archive directly:

```sh
oc-mirror --from .asoro/airgap-to-dir/mirror_seq1_000000.tar docker://registry.localhost:5000
```

@linfraze So I think this step can be removed and the rest of the procedure simplified like so:

1. Connect to your air-gapped environment, and make sure you are connected to the target registry and the target OCP cluster
2. From the air-gapped env, mirror the resources from the archive to the target registry by running the following command:
```
oc-mirror --from /path/to/mirror_seq1_000000.tar docker://registry.example.com:5000
```
 (replace `registry.example.com:5000` with your mirror registry)

As verification, the `oc-mirror` command should output something like:
```
Wrote release signatures to oc-mirror-workspace/results-1738075410
Writing image mapping to oc-mirror-workspace/results-1738075410/mapping.txt
Writing ICSP manifests to oc-mirror-workspace/results-1738075410
```

3. Keep track of the directory mentioned in the `oc-mirror` output
4.  Similar to https://github.com/redhat-developer/red-hat-developers-documentation-rhdh/pull/868#discussion_r1932188332, the user should create the ImageContentSourcePolicy resource on the cluster:
```
oc apply -f oc-mirror-workspace/results-1738075410/imageContentSourcePolicy.yaml
```

5. Deploy the Helm Chart:
```
CLUSTER_ROUTER_BASE=$(oc get route console -n openshift-console -o=jsonpath='{.spec.host}' | sed 's/^[^.]*\.//') helm install oc-mirror-workspace/results-1738070846/charts/redhat-developer-hub-{product-version}.tgz \
    --set global.clusterRouterBase="$CLUSTER_ROUTER_BASE"
```

rm3l (changes requested):
A few comments on the "fully disconnected environment" procedure

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -2,13 +2,13 @@
 // no assembly
 
 [id="proc-install-rhdh-airgapped-environment-ocp-helm_{context}"]
-= Installing {product} in an air-gapped environment with the Helm Chart
+= Installing {product} in an air-gapped environment with the Helm chart
```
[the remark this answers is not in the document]
linfraze: @rm3l Yeah, if we claim support for the other platforms, then we should document those air-gapped install procedures, too. 

I'm wondering if we should keep the air-gapped install title separate or if we should tuck an air-gapped install chapter into each of the installation guides. Either way, we can do that with a future PR. cc @Preeticp

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -2,137 +2,206 @@
...
+oc create -f oc-mirror-workspace/results-1738070846/ImageContentSourcePolicy.yaml
 ----
+
+. In your air-gapped environment, deploy the Helm chart to the namespace that you want to use by running the `helm install` command. For example:
 +
-[source,terminal,source,subs="attributes+"]
+[source,terminal,subs="attributes+"]
 ----
-podman pull registry.redhat.io/rhel9/postgresql-15:latest
+helm install rhdh ./mirror-archive/oc-mirror-workspace/src/charts/redhat-developer-hub-{product-version}.tgz --namespace <your-namespace> --create-namespace
```
linfraze, line 93: can these other parts of the command also be replaced with variables with definitions and examples? 
like `mirror-archive` or `redhat-developer-hub-{product-version}`?

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -2,137 +2,212 @@
...
 ====
-
-. Use the following command to verify that both images are present in the internal {ocp-short} registry:
 +
-[source,terminal]
+.Example output
+[source,terminal,subs="attributes+"]
 ----
-oc get imagestream -n {my-product-namespace}
+Creating archive /path/to/mirror-archive/mirror_seq1_000000.tar
```
linfraze, line 172: I have questions about the archive in general, and the `mirror-archive` and `mirror_seq1_000000.tar` examples and how to talk about them.

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -2,137 +2,212 @@
...
+* You have installed the {openshift-cli} on your workstation.
+* You are logged in to your target mirror registry and have permissions to push images to it. For more information, see link:https://docs.openshift.com/container-platform/4.17/disconnected/mirroring/installing-mirroring-disconnected.html#installation-adding-registry-pull-secret_installing-mirroring-disconnected[Configuring credentials that allow images to be mirrored].
+* You have installed the oc-mirror {openshift-cli} plugin, for more information see https://docs.openshift.com/container-platform/4.17/disconnected/mirroring/installing-mirroring-disconnected.html#installation-oc-mirror-installing-plugin_installing-mirroring-disconnected[Installing the oc-mirror OpenShift CLI plugin].
+* You have an account in https://developers.redhat.com/[{rhdeveloper-name}] portal.
+
+.Procedure
+. Create an `ImageSetConfiguration` file to specify the resources that you want to mirror. For example:
```
linfraze, line 129: TBH, I'm a little confused about what step 1 is for the fully disconnected procedure. Would love to chat in a huddle or see a demo.

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
rm3l: Yes, I think so.

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -2,203 +2,211 @@
...
+          - name: <chart-name> (3)
+            version: "<{product-very-short}-version>" (4)
 ----
-+
-[NOTE]
-====
-You can run the following commands to get the full host name of the {ocp-short} image registry, and then use the host name in a command to log in:
+<1> The name of the repository that you want to mirror, for example, `openshift-charts`.
+<2> The URL for the repository that you want to mirror, for example, `https://charts.openshift.io`.
+<3> The name of the chart that you want to mirror, for example, `redhat-developer-hub`.
```
linfraze, line 51: Helm chart that you want to mirror?

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -2,203 +2,211 @@
...
+    repositories:
+           - name: <repository-name> (1)
+        url: <repository-url> (2)
+        charts:
+          - name: <chart-name> (3)
+            version: "<{product-very-short}-version>" (4)
 ----
+<1> The name of the repository that you want to mirror, for example, `openshift-charts`.
+<2> The URL for the repository that you want to mirror, for example, `https://charts.openshift.io`.
+<3> The name of the chart that you want to mirror, for example, `redhat-developer-hub`.
```
linfraze, line 146: Helm chart

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
linfraze: @rm3l based on your other comment, are you suggesting that we replace`helm install rhdh ./mirror-archive/oc-mirror-workspace/src/charts/redhat-developer-hub-{product-version}.tgz --namespace <your-namespace> --create-namespace` with 
```
CLUSTER_ROUTER_BASE=$(oc get route console -n openshift-console -o=jsonpath='{.spec.host}' | sed 's/^[^.]*\.//') helm install oc-mirror-workspace/results-1738070846/charts/redhat-developer-hub-{product-version}.tgz \  
    --set global.clusterRouterBase="$CLUSTER_ROUTER_BASE"
``` 
???

If so, I have some follow-up Qs. :)

rm3l: Correct. I was thinking that `clusterRouterBase` could be determined and passed to the Helm CLI with one or two commands.

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -6,15 +6,19 @@
...
 
-If your network has access to the `registry.redhat.io` registry and the `charts.openshift.io` Helm Chart repository, you can deploy your {product} instance in your partially disconnected environment by mirroring the specified resources directly to the target registry.
+If your network has access to the `registry.redhat.io` registry and the `charts.openshift.io` Helm chart repository, you can deploy your {product} instance in your partially disconnected environment by mirroring the specified resources directly to the target registry.
 
 .Prerequisites
 
 * You have installed {ocp-brand-name} {ocp-version-min} or later.
-* You have access to the `registry.redhat.io`.
 * You have access to the `charts.openshift.io` Helm chart repository.
+* You have access to the `registry.redhat.io`.
```
linfraze, line 21: is this the same thing as the OCP image registry...? If so, I'll delete this bullet and specify this .io in the following bullet

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -113,7 +152,7 @@ helm install redhat-developer-hub./mirror-archive/oc-mirror-workspace/src/charts
 
 == Installing {product} on {ocp-short} in a fully disconnected environment with the Helm chart
 
-You can use the Helm chart to install {product} by mirroring specified resources and transferring them to your air-gapped environment without any connection to the internet.
+If your network has access to the registry through a bastion host or physical disk, you can use the Helm chart to install {product} by mirroring specified resources and transferring them to your air-gapped environment without any connection to the internet.
 
 .Prerequisites
```
linfraze, line 157: Need to double check that the prereqs section doesn't contain anything that doesn't apply to a fully disconnected environment and also isn't missing anything

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
rm3l: No, it is not the same. [`registry.redhat.io`](https://access.redhat.com/RegistryAuthentication) is the RH Ecosystem Catalog (a public authenticated registry for RH products), while the [OCP image registry](https://docs.openshift.com/container-platform/4.17/registry/index.html) is a registry deployed internally out of the box in each OCP cluster.

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -2,137 +2,312 @@
...
+
+`<results-directory>` :: Specifies the name of the image content source policy file, for example, `ImageContentSourcePolicy.yaml `.
+--
+. In your air-gapped environment, deploy the Helm chart to the namespace that you want to use by running the `helm install` command and `namespace` options. For example:
 +
-[source,terminal,source,subs="attributes+"]
+[source,terminal,subs="attributes+"]
 ----
-podman pull registry.redhat.io/rhdh/rhdh-hub-rhel9:{product-version}
+helm install <rhdh-instance> <workspace-directory>/<results-directory>/<charts-directory>/<archive-file> --namespace <your-namespace> --create-namespace
```
rm3l, line 122:
```suggestion
CLUSTER_ROUTER_BASE=$(oc get route console -n openshift-console -o=jsonpath='{.spec.host}' | sed 's/^[^.]*\.//')

helm install <rhdh-instance> <workspace-directory>/<results-directory>/<charts-directory>/<archive-file> --namespace <your-namespace> --create-namespace \
  --set global.clusterRouterBase="$CLUSTER_ROUTER_BASE"
```

I would suggest passing the `clusterRouterBase` parameter right when deploying the RHDH instance from the Helm Chart.
Otherwise, some things like the Home Page won't work if the user tries to access the RHDH instance from a browser.

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -2,137 +2,312 @@
...
+`<results-directory>` :: Specifies the name of your results directory, for example, `results-1738070846`.
+
+`<archive-file>` :: Specifies the name of the archive file containing the resources that you want to mirror, for example, `redhat-developer-hub-1.4.1.tgz`.
+
+`<your-namespace>` :: Specifies the namespace that you want to deploy the Helm chart to, for example, `{my-product-namespace}`.
+--
+
+.Verification
+. To verify that all resources are successfully deployed and running, check the status of your deployments and pods in the cluster.
+. View the router base URL by entering the following command:
```
rm3l, line 143: IMO, this step is not needed if the `clusterRouterBase` is passed in the previous `helm install` step.

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -2,137 +2,312 @@
...
++
+[source,terminal,subs="attributes+"]
 ----
-podman pull registry.redhat.io/rhel9/postgresql-15:latest
+CLUSTER_ROUTER_BASE=$(oc get route console -n openshift-console -o=jsonpath='{.spec.host}' | sed 's/^[^.]*\.//')
+helm install redhat-developer-hub./mirror-archive/oc-mirror-workspace/src/charts/redhat-developer-hub-1.4.1.tgz \
+    --set global.clusterRouterBase="$CLUSTER_ROUTER_BASE"
 ----
+. Ensure that the route matches the expected URL based on the router base of the cluster. If necessary, update the route to align with your expected configuration.
```
rm3l, line 151: Same point. This step is not needed if the `clusterRouterBase` is passed in the previous `helm install` step.

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -2,137 +2,312 @@
...
+
+== Installing {product} on {ocp-short} in a fully disconnected environment with the Helm chart
+
+If your network has access to the registry through a bastion host or physical disk, you can use the Helm chart to install {product} by mirroring specified resources and transferring them to your air-gapped environment without any connection to the internet.
```
rm3l, line 155:
```suggestion
If your network has access to the registry through a bastion host, you can use the Helm chart to install {product} by mirroring specified resources to disk and transferring them to your air-gapped environment without any connection to the internet.
```

I feel like this is not clear, as I am not sure how my network would have access to the registry through a physical disk. IMO, a physical disk is just a medium to transfer the resources over to the disconnected environment.

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
rm3l: I think maybe we could have different prerequisites:

1. on the user workstation
- You have access to the `registry.redhat.io`.
- You have access to the `charts.openshift.io` Helm chart repository.
- You have installed the {openshift-cli} on your workstation.
- You have installed the oc-mirror {openshift-cli} plugin, for more information see https://docs.openshift.com/container-platform/4.17/disconnected/mirroring/installing-mirroring-disconnected.html#installation-oc-mirror-installing-plugin_installing-mirroring-disconnected[Installing the oc-mirror OpenShift CLI plugin].
- You have an account in https://developers.redhat.com/[{rhdeveloper-name}] portal.

2. on the bastion host (which has access to the disconnected cluster and the target mirror registry)
- You have installed {ocp-brand-name} {ocp-version-min} or later.
- You have installed the {openshift-cli} on your workstation.
- You have access to a mirror registry that can be reached from the disconnected cluster, for example, the {ocp-brand-name} image registry. For more information about exposing the {ocp-short} image registry, see https://docs.redhat.com/en/documentation/openshift_container_platform/{ocp-version}/html-single/registry/index#securing-exposing-registry[Exposing the registry].
- You have installed the oc-mirror {openshift-cli} plugin, for more information see https://docs.openshift.com/container-platform/4.17/disconnected/mirroring/installing-mirroring-disconnected.html#installation-oc-mirror-installing-plugin_installing-mirroring-disconnected[Installing the oc-mirror OpenShift CLI plugin].

Because IMO, once the user has transferred all the necessary mirroring archive to their disconnected env, they no longer need to be connected to registry.redhat.io

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -2,137 +2,312 @@
...
+<3> The name of the Helm chart that you want to mirror, for example, `redhat-developer-hub`.
+<4> The version of {product} that you want to use, for example, `{product-version}`
+
+. Mirror the resources specified in the `ImageSetConfiguration` file by running the `oc-mirror` command. For example:
 +
-[source,terminal]
+[source,terminal,subs="attributes+"]
 ----
-podman push --remove-signatures registry.redhat.io/rhel9/postgresql-15:latest default-route-openshift-image-registry.<hostname>/<project_name>/postgresql-15:latest
+oc-mirror --config=<mirror-config-directory>/<mirror-configuration-file> <mirror-archive-directory>/<mirror-archive-file>
```
rm3l, line 192:
```suggestion
oc-mirror --config=<mirror-config-directory>/<mirror-configuration-file> <mirror-archive-directory>/
```

The user only needs to specify a path to a directory. `oc-mirror` will take care of creating an archive into that directory.

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -2,137 +2,312 @@
...
+--
+where:
+
+`<mirror-config-directory>` :: Specifies the location of your image set configuration file on your system, for example, `.user`.
+
+`<mirror-configuration-file>` :: Specifies the name of your image set configuration file, for example, `ImageSetConfiguration.yaml`.
+
+`<mirror-configuration-file>` :: Specifies the name of your mirror configuration yaml file, for example, `mirror-config.yaml`
+
+`<mirror-archive-directory>` :: Specifies the location of your `.tar` file containing the mirror archive, for example,`.user`.
```
rm3l, line 204:
```suggestion
`<mirror-archive-directory>` :: Specifies the location of your directory that will containing the mirror archive, for example,`.user`.
```

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -2,137 +2,312 @@
...
+
+`<mirror-archive-file>` :: Specifies the location and name of your `.tar` file containing the mirror archive, for example,`file://mirror-archive`.
```
rm3l, line 206:
```suggestion
```

Not needed, per my comment above.

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -2,137 +2,312 @@
...
++
+[source,terminal,subs="attributes+"]
 ----
-upstream:
-  postgresql:
-    image:
-      registry: ""
-      repository: postgresql-15
-      tag: latest
+helm install <rhdh-instance> <workspace-directory>/<results-directory>/<charts-directory>/<archive-file> --namespace <your-namespace> --create-namespace
```
rm3l, line 284:
```suggestion
CLUSTER_ROUTER_BASE=$(oc get route console -n openshift-console -o=jsonpath='{.spec.host}' | sed 's/^[^.]*\.//')

helm install <rhdh-instance> <workspace-directory>/<results-directory>/<charts-directory>/<archive-file> --namespace <your-namespace> --create-namespace \
  --set global.clusterRouterBase="$CLUSTER_ROUTER_BASE"
```

Same point as for the partially disconnected procedure.

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -2,137 +2,312 @@
...
+
+`<archive-file>` :: Specifies the name of the archive file containing the resources that you want to mirror, for example, `redhat-developer-hub-1.4.1.tgz`.
 
-. Install the {product} using Helm chart.
+`<your-namespace>` :: Specifies the namespace that you want to deploy the Helm chart to, for example, `{my-product-namespace}`.
+--
+
+.Verification
+. To verify that all resources are successfully deployed and running, check the status of your deployments and pods in the cluster.
+. View the router base URL by entering the following command:
```
rm3l, line 305: Same point as for the partially disconnected procedure. Not needed.

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -2,137 +2,312 @@
...
++
+[source,terminal,subs="attributes+"]
+----
+CLUSTER_ROUTER_BASE=$(oc get route console -n openshift-console -o=jsonpath='{.spec.host}' | sed 's/^[^.]*\.//')
+helm install redhat-developer-hub./mirror-archive/oc-mirror-workspace/src/charts/redhat-developer-hub-1.4.1.tgz \
+    --set global.clusterRouterBase="$CLUSTER_ROUTER_BASE"
+----
+. Ensure that the route matches the expected URL based on the router base of the cluster. If necessary, update the route to align with your expected configuration.
```
rm3l, line 313: Same point as for the partially disconnected procedure. Not needed.

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -2,137 +2,312 @@
...
+
+`<target-mirror-registry>` :: Specifies the location and name of your target mirror registry, for example,`docker://registry.example:5000`.
+--
++
 [NOTE]
 ====
-You can run the following commands to get the full host name of the {ocp-short} image registry, and then use the host name in a command to log in:
-
-[source,terminal]
+Running the `oc-mirror` command creates a local workspace containing the Helm chart and a `ImageContentSourcePolicy` (ICSP) manifest. The ICSP manifest contains an automatically-generated `imageContentSourcePolicy.yaml` file that you must apply against the cluster in a later step.
```
linfraze, line 77: @rm3l should we also mention the mirror archive here? 
```suggestion
Running the `oc-mirror` command creates a local workspace containing the mirror archive, the Helm chart, and a `ImageContentSourcePolicy` (ICSP) manifest. The ICSP manifest contains an automatically-generated `imageContentSourcePolicy.yaml` file that you must apply against the cluster in a later step.
```

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
linfraze: Will every fully disconnected user have a bastion host?

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
rm3l: The mirror archive is created only when we mirror to a directory. So it would make sense only for the fully disconnected environment use case.

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
rm3l: Yes, I think so. That's the definition of a fully disconnected environment. You cannot reach it directly from your workstation and the env does not have any access to the internet.

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
rm3l: This has been marked as resolved, but I'm still seeing the previous sentence:

image

WDYT about this?

rm3l:
@linfraze LGTM so far, except this review comment: https://github.com/redhat-developer/red-hat-developers-documentation-rhdh/pull/868#discussion_r1943059767
PTAL. Thanks.

Fortune-Ndlovu:
Reviewed and tested the doc update. Thanks @linfraze, @rm3l :rocket: 
/lgtm

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
linfraze: @rm3l ah, I accepted your change in my editor but it looks like it wasn't reflected here -- I definitely intend to make this change and will ensure that it is applied.

linfraze:
@Fortune-Ndlovu can you please update the `release-1.4` branch on your fork, please?

rm3l (approved):
/lgtm

rm3l:
> @Fortune-Ndlovu can you update the `release-1.4` branch on your fork, please?

This should also be in the `main` branch, for inclusion in 1.5.0

rm3l:
Just to make it clear that the procedure is OCP-specific.

/retitle Update the Helm procedure to use oc-mirror for air-gapped OCP environments

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -2,137 +2,284 @@
...
+CLUSTER_ROUTER_BASE=$(oc get route console -n openshift-console -o=jsonpath='{.spec.host}' | sed 's/^[^.]*\.//')
+
+helm install _<rhdh_instance>_ _<workspace_directory>_/_<results_directory>_/charts/_<archive_file>_ --namespace _<your_namespace>_ --create-namespace \
+  --set global.clusterRouterBase="$CLUSTER_ROUTER_BASE"
 ----
++
 --
+where:
+
+`<rhdh_instance>` :: Specifies the name of your {product} instance, for example, `my-rhdh`
```
pabel-rh, line 276: Missing a period at the end of the line.

Fortune-Ndlovu:
Closing this issue on release-1.4 branch and re-opened against the main branch here [#943](https://github.com/redhat-developer/red-hat-developers-documentation-rhdh/pull/943). The reason for the change to main is because this docs update introduces a breaking change, not a patch.

rm3l:
> Closing this issue on release-1.4 branch and re-opened against the main branch here [#943](https://github.com/redhat-developer/red-hat-developers-documentation-rhdh/pull/943). The reason for the change to main is because this docs update introduces a breaking change, not a patch.

Would have probably been better to rebase this PR branch onto `main` instead of opening a different PR..

linfraze:
> > Closing this issue on release-1.4 branch and re-opened against the main branch here [#943](https://github.com/redhat-developer/red-hat-developers-documentation-rhdh/pull/943). The reason for the change to main is because this docs update introduces a breaking change, not a patch.
> 
> Would have probably been better to rebase this PR branch onto `main` instead of opening a different PR..

Yes, I agree

linfraze (approved):
This PR has been co-authored and reviewed

modules/installation/proc-install-rhdh-airgapped-environment-ocp-helm.adoc
```diff
@@ -2,137 +2,284 @@
...
```
linfraze, line 276:
```suggestion
`<rhdh_instance>` :: Specifies the name of your {product} instance, for example, `my-rhdh`.
```

PR #875, opened 2025-01-24
Add compatibility matrix

jmagak:

**Version(s):** 1.4 

**Issue:** https://issues.redhat.com/browse/RHIDP-5070

### DIFF
--- a/artifacts/attributes.adoc
+++ b/artifacts/attributes.adoc
@@ -85,9 +85,9 @@
 :gke-short: GKE
 :gcp-brand-name: Google Cloud Platform
 :gcp-short: GCP
-:aro-brand-name: Azure Red Hat Openshift
+:aro-brand-name: Azure Red Hat OpenShift
 :aro-short: ARO
-:rosa-brand-name: Red Hat Openshift Service on AWS
+:rosa-brand-name: Red Hat OpenShift Service on AWS
 :rosa-short: ROSA
 
 // Release Notes
--- a/artifacts/attributes.adoc
+++ b/artifacts/attributes.adoc
@@ -85,6 +85,10 @@
 :gke-short: GKE
 :gcp-brand-name: Google Cloud Platform
 :gcp-short: GCP
+:aro-brand-name: Azure Red Hat Openshift
+:aro-short: ARO
+:rosa-brand-name: Red Hat Openshift Service on AWS
+:rosa-short: ROSA
 
 // Release Notes
 :rn-product-title: Release notes for Red Hat Developer Hub
--- a/modules/release-notes/ref-release-notes-compatibility-matrix.adoc
+++ b/modules/release-notes/ref-release-notes-compatibility-matrix.adoc
@@ -2,7 +2,7 @@
 [id="compatibility-matrix"]
 = Compatibility matrix
 
-The following table lists products and their versions compatible with {product-short} {product-version}:
+The following table lists products and their versions compatible with {product} {product-version}:
 
 .Platform Compatibility Matrix
 [cols="2,1", options="header"]
--- a/modules/release-notes/ref-release-notes-compatibility-matrix.adoc
+++ b/modules/release-notes/ref-release-notes-compatibility-matrix.adoc
@@ -5,11 +5,9 @@
 The following table lists products and their versions compatible with {product-short} {product-version}:
 
 .Platform Compatibility Matrix
-[cols="1,2,1", options="header"]
+[cols="2,1", options="header"]
 |===
-| {product-short} | Platform | {product-custom-resource-type} Version
-
-| `{product-version}`
+| Platform | {product-custom-resource-type} Version
 |
 - {ocp-brand-name} `{ocp-version-min}`-`{ocp-version}`
 
--- a/modules/release-notes/ref-release-notes-compatibility-matrix.adoc
+++ b/modules/release-notes/ref-release-notes-compatibility-matrix.adoc
@@ -1,0 +1,41 @@
+:_content-type: REFERENCE
+[id="compatibility-matrix"]
+= Compatibility matrix
+
+The following table lists products and their versions compatible with {product-short} {product-version}:
+
+.Platform Compatibility Matrix
+[cols="1,2,1", options="header"]
+|===
+| {product-short} | Platform | {product-custom-resource-type} Version
+
+| `{product-version}`
+|
+- {ocp-brand-name} `{ocp-version-min}`-`{ocp-version}`
+
+- {eks-brand-name} ({eks-short})
+
+- {aks-brand-name} ({aks-short})
+
+- {aro-brand-name} ({aro-short})
+
+- {rosa-brand-name} ({rosa-short})
+
+- {gke-brand-name} ({gke-short})
+| `{product-backstage-version}`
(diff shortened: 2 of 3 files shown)


modules/release-notes/ref-release-notes-compatibility-matrix.adoc
```diff
@@ -0,0 +1,41 @@
...
+
+- {aro-brand-name} ({aro-short})
+
+- {rosa-brand-name} ({rosa-short})
+
+- {gke-brand-name} ({gke-short})
+| `{product-backstage-version}`
+| `1.3`
+|
+- {ocp-brand-name} `4.14`-`4.16`
```
nickboldt, line 28: Please use attributes for OCP min/max values: https://github.com/redhat-developer/red-hat-developers-documentation-rhdh/blob/release-1.4/artifacts/attributes.adoc#L42-L43 

```suggestion
- {ocp-brand-name} `{ocp-version-min}`-`{ocp-version}`
```

modules/release-notes/ref-release-notes-compatibility-matrix.adoc
```diff
@@ -0,0 +1,41 @@
...
+- {eks-brand-name} ({eks-short})
+
+- {aks-brand-name} ({aks-short})
+
+- {aro-brand-name} ({aro-short})
+
+- {rosa-brand-name} ({rosa-short})
+
+- {gke-brand-name} ({gke-short})
+| `1.29`
```
nickboldt, line 39: See https://github.com/redhat-developer/red-hat-developers-documentation-rhdh/blob/release-1.3/artifacts/attributes.adoc#L16

```suggestion
| `1.29.2`
```

modules/release-notes/ref-release-notes-compatibility-matrix.adoc
```diff
@@ -0,0 +1,41 @@
...
+
+| `{product-version}`
+|
+- {ocp-brand-name} `{ocp-version-min}`-`{ocp-version}`
+
+- {eks-brand-name} ({eks-short})
+
+- {aks-brand-name} ({aks-short})
+
+- {aro-brand-name} ({aro-short})
```
nickboldt, line 20: aro and rosa are not yet in the attributes.adoc file:

https://github.com/redhat-developer/red-hat-developers-documentation-rhdh/blob/release-1.4/artifacts/attributes.adoc#L73-L87

So you'll want to add those before this will render correctly.

The preview looks like this:

image

modules/release-notes/ref-release-notes-compatibility-matrix.adoc
```diff
@@ -0,0 +1,41 @@
...
+
+- {rosa-brand-name} ({rosa-short})
+
+- {gke-brand-name} ({gke-short})
+| `{product-backstage-version}`
+| `1.3`
```
nickboldt, line 26: Do we want/need a 1.3 table in the 1.4 docs? Or should this be added instead in the `release-1.3` branch and published to the 1.3 docs?

nickboldt:
Looks OK now 
image

Though that first column is sorta redundant now as we're only talking about the current release version, rather than listing multiple versions.

You could replace the need to maintain this table entiurely with a link to the `Supported Platforms` table at https://access.redhat.com/support/policy/updates/developerhub 

But if @reestr and @christophe-f  are happy with this duplication, then so am I.

nickboldt (approved):

linfraze:
I like the idea of linking to the the Supported Platforms table as the single source of truth, as @nickboldt suggested. Then we don't have to worry about duplicating existing information, maintaining multiple sources, or getting out of sync. But if there is a reason why it makes sense to embed the duplicate info directly in the RNs, then we can do the needful.

Gerry-Forde:
@jmagak For some reason, this link to the Compatibility Matrix links to a PR, can you review this please:
Screenshot 2025-01-27 at 17 47 01

nickboldt:
> having a bulleted list of platforms

... which you would then have to update every quarter if new platforms/versions are added. 

https://access.redhat.com/support/policy/updates/developerhub is already a list... inside a table.

artifacts/attributes.adoc
```diff
@@ -85,6 +85,10 @@
 :gke-short: GKE
 :gcp-brand-name: Google Cloud Platform
 :gcp-short: GCP
+:aro-brand-name: Azure Red Hat Openshift
```
hmanwani-rh, line 88:
```suggestion
:aro-brand-name: Azure Red Hat OpenShift
```

artifacts/attributes.adoc
```diff
@@ -85,6 +85,10 @@
...
+:aro-short: ARO
+:rosa-brand-name: Red Hat Openshift Service on AWS
```
hmanwani-rh, line 90:
```suggestion
:rosa-brand-name: Red Hat OpenShift Service on AWS
```

modules/release-notes/ref-release-notes-compatibility-matrix.adoc
```diff
@@ -0,0 +1,25 @@
+:_content-type: REFERENCE
+[id="compatibility-matrix"]
+= Compatibility matrix
+
+The following table lists products and their versions compatible with {product-short} {product-version}:
```
hmanwani-rh, line 5:
```suggestion
The following table lists products and their versions compatible with {product} {product-version}:
```

hmanwani-rh (commented):
@jmagak I’ve added a few minor comments. However, I recall that we planned to include a table for the supported versions in the release notes. For example, the 1.5 release notes would exclude version 1.3 and only include 1.5 and 1.4, which I think we can maintain indicating the relevant supported versions in a release. Additionally, how about we share the lifecycle page link in the "Additional Resources" section, similar to how it’s done in the "Supported Platforms" section?

Let me know your thoughts!

nickboldt:
> Let me know your thoughts!

My $0.02 cents remains "why duplicate content when you can just link to the lifecycle page as the single source of truth?"

hmanwani-rh:
> > Let me know your thoughts!
> 
> My $0.02 cents remains "why duplicate content when you can just link to the lifecycle page as the single source of truth?"
> 
> Also having 1.3 content in the 1.4 guide ... what value does that provide? is the idea to provide people with immediate access to seeing what has changed between 1.3 and 1.4 in terms of supported platforms/versions? If so that is again already in the lifecyle doc.

@nickboldt The Lifecycle page currently doesn’t include the compatible Backstage versions, which I think would be helpful to show in our documentation. What are your thoughts on this? 
Additionally, we may also need to include the tested hyperscaler versions.

nickboldt:
Backstage version is already documented in https://github.com/redhat-developer/red-hat-developers-documentation-rhdh/blob/release-1.4/titles/rel-notes-rhdh/title-rhdh-release-notes.adoc

image

nickboldt:
> we may also need to include the tested hyperscaler versions

Sounds like something to add for 1.5. We don't have it for 1.2-1.4, so why not add this New Feature in the upcoming release instead of the maintenance ones?

jmagak:
Instead of a Compatibility Matrix, we are adding an Additional Resources section that directs users to the lifecycle page. Hence, we are closing this PR.

PR #879, opened 2025-01-27
RHIDP:3206: added attributes subs to the source block

pabel-rh:

**Version(s):** main, 1.4.x, 1.3.x
**Issue:** [RHIDP-3206](https://issues.redhat.com/browse/RHIDP-3206)

### DIFF
--- a/artifacts/rhdh-plugins-reference/aap-backend/aap-backend-install-dynamic-plugin.adoc
+++ b/artifacts/rhdh-plugins-reference/aap-backend/aap-backend-install-dynamic-plugin.adoc
@@ -1,7 +1,7 @@
 .Installation
 The AAP backend plugin is pre-loaded in {product-short} with basic configuration properties. To enable it, set the `disabled` property to `false` as follows:
 
-[source,yaml]
+[source,yaml,subs="+attributes"]
 ----
 global: 
   dynamic: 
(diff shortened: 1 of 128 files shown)


linfraze (approved):
the code looks right, but if you know where to find a preview of an example of an attribute being rendered in a code block in one of these changed files, please drop a link in the PR so we can easily see what the output looks like. 

/lgtm

linfraze:
@pabel-rh reminder to cherry-pick only to 1.3 since earlier version are EOL.

themr0c:
The addition of the `subs="+attributes"` statement is not necessary when the code block does not contain any attribute.

pabel-rh:
I had to commit again to fix a few merge conflicts.
@linfraze , here is a preview link for one example of the usage of attributes within the source code block:

https://redhat-developer.github.io/red-hat-developers-documentation-rhdh/pr-879/authorization/#proc-sending-requests-to-the-rbac-rest-api-by-using-curl_title-authorization

@themr0c, I had gone ahead with Nick's alternate approach which he mentioned in the comments section, as that way, it someone in the future does edit one of the source code blocks which currently does not have any attributes, it just makes it easier.

nickboldt:
The missing piece here is as described in the JIRA:
> add a grep check in the build.sh (and/or build-ccutil.sh) that will fail if any
> `[source]`
> blocks are found which do NOT declare the
> `subs="attributes+"`
> rendering instruction.

So while adding all the instuctions is half the battle, anyone who adds new code snippets may fall into the same rendering trap in future unless we ALSO implement something in the build.sh or build-ccutil.sh (or some new PR check) which will ensure ALL code blocks can render attributes (even if they don't contain any). @themr0c WDYT?

PR #891, opened 2025-01-29
RHIDP-5489 Enabling authentication with your authentication provider

themr0c:
Enabling authentication with your custom authentication provider.

Preview: https://redhat-developer.github.io/red-hat-developers-documentation-rhdh/pr-891/authentication/#enabling-authentication-with-your-custom-authentication-provider

**Version(s):** 1.5

**Issue:** https://issues.redhat.com/browse/RHIDP-5489


assemblies/assembly-enabling-authentication.adoc
```diff
@@ -61,3 +61,5 @@ include::assembly-authenticating-with-github.adoc[leveloffset=+1]
 
 include::assembly-authenticating-with-microsoft-azure.adoc[leveloffset=+1]
 
+
+include::modules/authentication/proc-enabling-authentication-with-your-custom-authentication-provider.adoc[leveloffset=+1]
```
themr0c, line 65: @gashcrumb Do you agree that it is something we want to document as part of the _Authentication_ title?

modules/authentication/proc-enabling-authentication-with-your-custom-authentication-provider.adoc
```diff
@@ -0,0 +1,212 @@
...
+          module: PluginRoot
+          importName: SignInPage
+----
+====
+--
+
+. link:{installing-and-viewing-dynamic-plugins-url}#assembly-third-party-plugins[Export, package, and install your custom plugins].
+. Add your custom plugins configuration to `dynamic-plugins.yaml` in {product-short}.
++
+.`dynamic-plugins.yaml` excerpt for LDAP catalog support
```
themr0c, line 112: @gashcrumb Does the `dynamic-plugins.yaml` file content need to be that long? I would assume that if the plugins has defined default values, it is sufficient to enable the plugin. Is it correct?

assemblies/assembly-enabling-authentication.adoc
gashcrumb: Yes, I think so.  It's an alternate way of setting up authentication in RHDH, where our opinionated static authentication provider module is disabled.  This allows the user to install a suite of dynamic plugins that address their specific authentication concerns.

modules/authentication/proc-enabling-authentication-with-your-custom-authentication-provider.adoc
gashcrumb: Maybe these LDAP plugins aren't a great example, most of the config here is concerned with getting these plugins to talk to LDAP.  It's probably less important than the touch points where the user is required to integrate either configuration or code to connect authentication providers to the RHDH app.  I have just finished up another example that I could walk you through if you like [here](https://github.com/gashcrumb/dynamic-plugins-multiple-auth-modules).

modules/authentication/proc-enabling-authentication-with-your-custom-authentication-provider.adoc
```diff
@@ -0,0 +1,231 @@
+[id="enabling-authentication-with-your-custom-authentication-provider"]
+= Enabling authentication with your custom authentication provider
+
+To authenticate users with your custom authentication provider, create, install and configure your custom authentication backend and frontend plugins in {product}.
```
themr0c, line 4:
```suggestion
You can extend {product} with your custom authentication provider by using your custom authentication backend and frontend plugins.
To authenticate users with your custom authentication provider, create, install and configure your custom authentication backend and frontend plugins in {product}.
```

assemblies/assembly-enabling-authentication.adoc
themr0c: Perfect.

modules/authentication/proc-enabling-authentication-with-your-custom-authentication-provider.adoc
themr0c: At first sight this example is maybe even more complex, because we have 4 plugins.

Would an example plugin defining static users be sufficient for the demonstration purpose? Or is it a requirement to delegate authentication to an external service?

themr0c:
For a complex example, we could refer to a Git repository in the OpenShift GitHUb namespace, such as done for pipelines: https://docs.openshift.com/pipelines/1.17/create/creating-applications-with-cicd-pipelines.html

themr0c:
Short term: integrate https://github.com/redhat-developer/rhdh/pull/2446/files without the link to the examples.

PR #898, opened 2025-01-30
Added missing parameter in auth set up

jmagak:
Adds the `prompt: auto` config which is not set in the current RHDH docs for RHBK auth provider set up [here](https://docs.redhat.com/en/documentation/red_hat_developer_hub/1.4/html/authentication/assembly-authenticating-with-rhbk#enabling-authentication-with-rhbk).

**Version(s):** 1.4

**Issue:** https://issues.redhat.com/browse/RHIDP-5681


JessicaJHee (changes requested):
Thank you for this! I have a few small comments.

It would also be nice to add a description [here](https://github.com/redhat-developer/red-hat-developers-documentation-rhdh/blob/8e8ac41366e9a1bf09ea4fc37b139daff78e6835/modules/authentication/proc-enabling-authentication-with-rhbk.adoc?plain=1#L59) explaining this config:

Something like: 
`prompt: auto`: To allow the identity provider to automatically determine whether to prompt for credentials or skip the login redirect if you have an active SSO session. 

**Some more useful information to consider including:**

Without setting this value, the provider will default to `prompt:  none`, which assumes that you are already logged on in the IDP and will reject the sign-in request without an active session.

`prompt: login`: will force the identity provider to always require credentials from the user.

modules/authentication/proc-enabling-authentication-with-rhbk.adoc
```diff
@@ -45,7 +45,9 @@ auth:
         metadataUrl: ${AUTH_OIDC_METADATA_URL}
         clientId: ${AUTH_OIDC_CLIENT_ID}
         clientSecret: ${AUTH_OIDC_CLIENT_SECRET}
+        prompt: auto #add this config
 signInPage: oidc
+dangerouslyAllowSignInWithoutUserInCatalog: true
```
JessicaJHee, line 50: Please remove this field as it is not mandatory (and also not safe for production)

modules/authentication/proc-enabling-authentication-with-rhbk.adoc
```diff
@@ -45,7 +45,9 @@ auth:
         metadataUrl: ${AUTH_OIDC_METADATA_URL}
         clientId: ${AUTH_OIDC_CLIENT_ID}
         clientSecret: ${AUTH_OIDC_CLIENT_SECRET}
+        prompt: auto #add this config
```
JessicaJHee, line 48:
```suggestion
        prompt: auto
```

modules/authentication/proc-enabling-authentication-with-rhbk.adoc
```diff
@@ -78,6 +85,7 @@ auth:
...
```
JessicaJHee, line 88:
```suggestion
        prompt: auto
```

modules/authentication/proc-enabling-authentication-with-rhbk.adoc
JessicaJHee, line 50:
```suggestion
```

JessicaJHee (changes requested):
I've added a few more comments, thank you!

modules/authentication/proc-enabling-authentication-with-rhbk.adoc
```diff
@@ -35,7 +35,7 @@ Save the value for the next step:
 +
 --
 .`{my-app-config-file}` fragment with mandatory fields to enable authentication with {rhbk}
-[source,yaml]
+[source,yaml,subs="+attributes,+quotes"]
```
themr0c, line 38: Consider omitting this substitutions statement, which is confusing and misleading:

* In this example, if you enable the `attributes` substitutions, you assert that `{AUTH_OIDC_METADATA_URL}`, `AUTH_OIDC_CLIENT_ID}`, and `{AUTH_OIDC_CLIENT_SECRET}` are attributes, which is not the case.

* As there is no user substituted value, the `quotes` substitution is unused, and not necessary.

JessicaJHee (approved):
LGTM, thank you for these changes!

hmanwani-rh:
@jmagak Please rebase the PR

PR #913, opened 2025-02-06
RHIDP-5865: Replaced hyphens with underscore in the attributes

pabel-rh:

**Version(s):**
main, 1.5, 1.4.x, 1.3.x
**Issue:**
[RHIDP-5865](https://issues.redhat.com/browse/RHIDP-5865)


linfraze (commented):
If we are going to use attributes in this way, let's make sure we are properly differentiating between a file/object **variable** and an **example**. Each serves a specific purpose. 

A variable: 

- Explains what the file/object _is_ so the user knows what to replace it with
- Shows the user which file/object names are replaceable in a code sample vs. which are verbatim
- Follows CCS supplementary style (italics, underscores)

An example file/object: 
- Shows the user a sample file/object name and the expected format (such as `my-app-config.yaml`)

pabel-rh:
I see your point, @linfraze. In that case, I believe there's no need to implement this change since these attributes are for examples, instead of variables. :) I'm sorry I misunderstood previously and created this request. I will close this PR. :)

PR #935, opened 2025-02-13
RHIDP-5494: Doc air-gapped installation using Operator

linfraze:

**Version(s):**
1.5

**Issue:**
RHIDP-5494

**Preview:**
https://redhat-developer.github.io/red-hat-developers-documentation-rhdh/pr-935/install-rhdh-air-gapped/#proc-install-rhdh-airgapped-environment-ocp-operator_title-install-rhdh-air-grapped

### DIFF
--- a/modules/installation/artifacts
+++ b/modules/installation/artifacts
@@ -1,1 +1,0 @@
-../../artifacts
--- a/modules/installation/proc-install-rhdh-airgapped-environment-ocp-operator.adoc
+++ b/modules/installation/proc-install-rhdh-airgapped-environment-ocp-operator.adoc
@@ -153,4 +153,10 @@ The script can take several minutes to complete as it automatically installs the
 kubectl -n rhdh-operator get pods
 ----
 
-include::../../artifacts/snip-rhdh-install-operator-next-steps.adoc[]
+.Next steps
+* Use the Operator to create a {product} instance on a supported platform. For more information, see the following documentation for the platform that you want to use:
+** link:https://docs.redhat.com/en/documentation/red_hat_developer_hub/1.4/html/installing_red_hat_developer_hub_on_openshift_container_platform/assembly-install-rhdh-ocp-operator[Installing {product} on {ocp-short} with the Operator]
+** link:https://docs.redhat.com/en/documentation/red_hat_developer_hub/1.4/html/installing_red_hat_developer_hub_on_amazon_elastic_kubernetes_service/proc-rhdh-deploy-eks-operator_title-install-rhdh-eks[Installing {product-short} on {eks-short} with the Operator]
+** link:https://docs.redhat.com/en/documentation/red_hat_developer_hub/1.4/html/installing_red_hat_developer_hub_on_microsoft_azure_kubernetes_service/proc-rhdh-deploy-aks-operator_title-install-rhdh-aks[Installing {product-short} on {aks-short} with the Operator]
+** link:https://docs.redhat.com/en/documentation/red_hat_developer_hub/1.4/html/installing_red_hat_developer_hub_on_openshift_dedicated_on_google_cloud_platform/proc-install-rhdh-osd-gcp-operator_title-install-rhdh-osd-gcp[Installing {product-short} on {gcp-short} with the Operator]
+** link:https://docs.redhat.com/en/documentation/red_hat_developer_hub/1.4/html/installing_red_hat_developer_hub_on_google_kubernetes_engine/proc-rhdh-deploy-gke-operator.adoc_title-install-rhdh-gke#proc-deploy-rhdh-instance-gke.adoc_title-install-rhdh-gke[Deploying {product-short} on {gke-short} with the Operator]


linfraze:
I will modularize this content after the SME review is completed

modules/installation/proc-install-rhdh-airgapped-environment-ocp-operator.adoc
```diff
@@ -5,68 +5,159 @@
...
+== Installing {product} on {ocp-short} in a fully disconnected environment with the Operator
+
+If your network has access to the registry through a bastion host or physical disk, you can use the Operator to install {product} by mirroring specified resources and transferring them to your air-gapped environment without any connection to the internet.
+
+.Prerequisites
+
+* You have set up your disconnected environment.
+** You have mirrored all of the required images to disk.
+** You have manually transferred the mirror folder to the network of the disconnected mirror registry.
+** You have mirrored the images from disk to the target mirror registry in your disconnected environment.
```
rm3l, line 84: Same point: this will be done by the script.

modules/installation/proc-install-rhdh-airgapped-environment-ocp-operator.adoc
```diff
@@ -5,68 +5,159 @@
...
+** You have installed the Operator in your disconnected environment.
```
rm3l, line 85: Same point: this is what the script ultimately does when it is done mirroring from disk to the target mirror registry.

modules/installation/proc-install-rhdh-airgapped-environment-ocp-operator.adoc
```diff
@@ -5,68 +5,159 @@
...
+* You have set up your workstation.
+** You have an active `oc registry` session to the `registry.redhat.io` {company-name} Ecosystem Catalog. For more information, see link:https://access.redhat.com/RegistryAuthentication[{company-name} Container Registry Authentication].
```
rm3l, line 87: Same point as in one of previous comments.. User can use `oc registry`, `podman login` or `skopeo login`.

modules/installation/proc-install-rhdh-airgapped-environment-ocp-operator.adoc
```diff
@@ -5,68 +5,159 @@
...
++
+[source,terminal,subs="attributes+"]
+----
+curl -sSLO https://raw.githubusercontent.com/redhat-developer/rhdh-operator/refs/heads/release-{product-version}/.rhdh/scripts/prepare-restricted-environment.sh
+----
++
+. Run the mirroring script by running the `bash` command with the appropriate set of options:
++
+[source,terminal,subs="+quotes,+attributes"]
+----
```
rm3l, line 102: The command you mentioned (using `--ci-index true`) is for CI testing purposes, and should not be part of the product docs. 

Instead, this is what the end-user needs to run to mirror the images to disk:

```
bash prepare-restricted-environment.sh \
    --to-dir _<my_pulled_image_location>_
```

modules/installation/proc-install-rhdh-airgapped-environment-ocp-operator.adoc
```diff
@@ -5,68 +5,159 @@
...
+    --index-image "quay.io/rhdh/iib:next-v4.18-x86_64" \
+    --ci-index true \
+    --to-dir _<my_pulled_image_location>_ \
+    [--filter-versions '*'] \
+    [--use-oc-mirror true]
+----
++
+where
+
+_<my_pulled_image_location>_ :: Specifies the directory where you want to pull all of the necessary images with the `--to-dir` option, for example, my.registry.example.com/namespace
```
rm3l, line 113:
```suggestion
_<my_pulled_image_location>_ :: Specifies the absolute path to a directory where you want to pull all of the necessary images with the `--to-dir` option, for example, /home/user/rhdh-operator-mirror-dir
```

`my.registry.example.com/namespace` sounds more like a registry path, not a directory. And I think we should mention that the path needs to be absolute at the moment.

modules/installation/proc-install-rhdh-airgapped-environment-ocp-operator.adoc
```diff
@@ -5,68 +5,159 @@
...
+[source,terminal,subs="+quotes,+attributes"]
+----
+kubectl -n rhdh-operator get pods
+----
+
+== Installing {product} on {ocp-short} in a fully disconnected environment with the Operator
+
+If your network has access to the registry through a bastion host or physical disk, you can use the Operator to install {product} by mirroring specified resources and transferring them to your air-gapped environment without any connection to the internet.
+
+.Prerequisites
```
rm3l, line 79: There are additional prerequisites (common to both partially and fully disconnected environments) mentioned in https://raw.githubusercontent.com/redhat-developer/rhdh-operator/refs/heads/main/.rhdh/docs/airgap.adoc:

> * GNU sed. See [GNU sed](https://www.gnu.org/software/sed/).
> 
> * (Optional) oc-mirror (the OpenShift Client mirror plugin) if you want to mirror using oc-mirror on OpenShift. See [Installing the oc-mirror OpenShift CLI plugin](https://docs.redhat.com/en/documentation/openshift_container_platform/4.17/html-single/disconnected_environments/index#installation-oc-mirror-installing-plugin_about-installing-oc-mirror-v2).
> 
> * Skopeo 1.17+. See [Installing Skopeo](https://github.com/containers/skopeo/blob/main/install.md).
> 
> * umoci. See [Install](https://github.com/opencontainers/umoci#install).
> 
> * yq v4.44+. See [Install](https://github.com/mikefarah/yq#install).
> 
> * (Required on non-OpenShift clusters) The cluster needs to have the Operator Lifecycle Manager installed and already working in disconnected environments. See [Installing OLM](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/doc/install/install.md#installing-olm).
> 
> * (Required on non-OpenShift clusters) A mirror registry reachable from the cluster.

modules/installation/proc-install-rhdh-airgapped-environment-ocp-operator.adoc
```diff
@@ -5,68 +5,159 @@
...
+
+_<my_pulled_image_location>_ :: Specifies the directory where you want to pull all of the necessary images with the `--to-dir` option, for example, my.registry.example.com/namespace
++
+[NOTE]
+====
+The script can take several minutes to complete as it copies multiple images to the mirror registry.
+====
++
+. Transfer the directory specified by the `--to-dir` option to your disconnected environment.
+. From a machine in your disconnected environment that has access to both the cluster and the target mirror registry, download the mirroring script from disk by running the following command:
```
rm3l, line 121: It is not clear from https://raw.githubusercontent.com/redhat-developer/rhdh-operator/refs/heads/main/.rhdh/docs/airgap.adoc, but there is no need to redownload the script again. In the mirror directory transferred by the user, the script will be there, named as `install.sh`.

modules/installation/proc-install-rhdh-airgapped-environment-ocp-operator.adoc
```diff
@@ -5,68 +5,159 @@
...
++
+[source,terminal,subs="attributes+"]
+----
+curl -sSLO https://raw.githubusercontent.com/redhat-developer/rhdh-operator/refs/heads/release-{product-version}/.rhdh/scripts/prepare-restricted-environment.sh
+----
++
+. Run the mirroring script by running the `bash` command with the appropriate set of options:
++
+[source,terminal,subs="+quotes,+attributes"]
+----
```
rm3l, line 131: Based on my previous comment, the command could be:

```
bash _<my_pulled_image_location>_/install.sh  \
    --from-dir _<my_pulled_image_location>_ \
    [--to-registry _<my.registry.example.com>_] \
    [--use-oc-mirror true]
```

modules/installation/proc-install-rhdh-airgapped-environment-ocp-operator.adoc
```diff
@@ -5,68 +5,159 @@
...
+    [--use-oc-mirror true]
+----
++
+where
+
+_<my_pulled_image_location>_ :: Specifies the directory where you want to pull all of the necessary images with the `--to-dir` option.
+
+_<my.registry.example.com>_ :: Specifies the URL for the target mirror registry where you want to mirror the images.
+
+_<namespace>_ :: Specifies the target namespace where you want to mirror the images.
```
rm3l, line 144:
```suggestion
```

I think we can remove this part, as it is confusing with a Kubernetes namespace. It is actually supposed to be part of the overall registry URL, but it depends on how each mirror registry handles that. It might not be needed on some registries.

rm3l (changes requested):

modules/installation/proc-install-rhdh-airgapped-environment-ocp-operator.adoc
```diff
@@ -5,68 +5,159 @@
 [id="proc-install-rhdh-airgapped-environment-ocp-operator_{context}"]
 = Installing {product} in an air-gapped environment with the Operator
 
+You can install {product} in a fully disconnected or partially disconnected environment using the {product} Operator. For a list of supported platforms, see the link:https://access.redhat.com/support/policy/updates/developerhub[{product} Life Cycle page].
+
+== Installing {product} on {ocp-short} in a partially disconnected environment with the Operator
```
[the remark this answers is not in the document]
linfraze: Good catch - The OCP reference is not supposed to be there and was removed from the main title.

modules/installation/proc-install-rhdh-airgapped-environment-ocp-operator.adoc
```diff
@@ -5,68 +5,159 @@
...
-====
-
-* If you prefer to create your own mirror registry, see link:https://docs.redhat.com/en/documentation/openshift_container_platform/{ocp-version}/html-single/disconnected_installation_mirroring/index#installing-mirroring-creating-registry[Creating a mirror registry with mirror registry for Red Hat OpenShift].
-
-* If you do not already have a mirror registry, you can use the helper script to create one for you and you need the following additional prerequisites:
-+
-** The cURL package is installed. For {rhel}, the curl command is available by installing the curl package. To use curl for other platforms, see the link:https://curl.se/[cURL website].
-** The `htpasswd` command is available. For {rhel}, the `htpasswd` command is available by installing the `httpd-tools` package.
+* You have an active `oc registry` session to the `registry.redhat.io` {company-name} Ecosystem Catalog. For more information, see link:https://access.redhat.com/RegistryAuthentication[{company-name} Container Registry Authentication].
```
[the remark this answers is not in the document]
linfraze: I think the second `podman` should be `skopeo` so will apply that change

modules/installation/proc-install-rhdh-airgapped-environment-ocp-operator.adoc
```diff
@@ -5,68 +5,159 @@
...
+----
+
+== Installing {product} on {ocp-short} in a fully disconnected environment with the Operator
+
+If your network has access to the registry through a bastion host or physical disk, you can use the Operator to install {product} by mirroring specified resources and transferring them to your air-gapped environment without any connection to the internet.
+
+.Prerequisites
+
+* You have set up your disconnected environment.
+** You have mirrored all of the required images to disk.
```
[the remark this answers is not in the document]
linfraze: Ah, good point. I pulled this from the upstream doc, btw.

modules/installation/proc-install-rhdh-airgapped-environment-ocp-operator.adoc
```diff
@@ -5,68 +5,141 @@
...
+----
+curl -sSLO https://raw.githubusercontent.com/redhat-developer/rhdh-operator/refs/heads/release-{product-version}/.rhdh/scripts/prepare-restricted-environment.sh
+----
++
+. Run the mirroring script by running the `bash` command with the appropriate set of options:
++
+[source,terminal,subs="+quotes,+attributes"]
+----
+bash prepare-restricted-environment.sh \ (1)
+  [--to-registry _<my.registry.example.com>_]] \ (2)
```
rm3l, line 52:
```suggestion
  --filter-versions "{product-version}" \
  [--to-registry _<my.registry.example.com>_] \ (2)
```

I realized that the default version installed is 1.4 in the script. The docs need to reflect the right version that the user will use (even if 1.5 is currently not GA yet).

modules/installation/proc-install-rhdh-airgapped-environment-ocp-operator.adoc
```diff
@@ -5,68 +5,141 @@
...
-
-# if you already have a target mirror registry
-# use the following example:
 bash prepare-restricted-environment.sh \
-   --prod_operator_index "registry.redhat.io/redhat/redhat-operator-index:v{ocp-version}" \
-   --prod_operator_package_name "rhdh" \
-   --prod_operator_bundle_name "rhdh-operator" \
-   --prod_operator_version "v{product-bundle-version}" \
-   --use_existing_mirror_registry "my_registry"
+    --to-dir _<my_pulled_image_location>_
```
rm3l, line 101:
```suggestion
    --filter-versions "{product-version}" \
    --to-dir _<my_pulled_image_location>_ \
    [--use-oc-mirror true]
```

modules/installation/proc-install-rhdh-airgapped-environment-ocp-operator.adoc
```diff
@@ -5,68 +5,141 @@
...
+.Verification
+* If you are using {ocp-brand-name}, the {product} Operator is in the *Installed Operators* list in the web console.
+* If you are using a supported Kubernetes platform, you can check the list of pods running in the `rhdh-operator` namespace by running the following command in your terminal:
 +
-** The cURL package is installed. For {rhel}, the curl command is available by installing the curl package. To use curl for other platforms, see the link:https://curl.se/[cURL website].
-** The `htpasswd` command is available. For {rhel}, the `htpasswd` command is available by installing the `httpd-tools` package.
+[source,terminal,subs="+quotes,+attributes"]
+----
+kubectl -n rhdh-operator get pods
+----
```
rm3l, line 70: The helper script installs the Operator only (not RHDH). It is then up to the user to install RHDH by creating a CR. Since the goal of this page is ultimately "Installing Red Hat Developer Hub in an air-gapped environment", do you want to add further instructions on how to install RHDH once the Operator is up and running (or link to other procedures)?

modules/installation/proc-install-rhdh-airgapped-environment-ocp-operator.adoc
linfraze: You're right. We'll actually use _<product_version>_ in the code to indicate replaceable text where the user can insert whichever product version they want to use and we'll use the attribute in the variable definition/example.

modules/installation/proc-install-rhdh-airgapped-environment-ocp-operator.adoc
linfraze: @rm3l This is a Q I had - is the rest of the disconnected procedure the same as installing RHDH (on whichever supported platform the customer is using) on a connected network? If yes, then we can link to those existing procedures in a **Next steps** section. If the disconnected process requires additional/different steps, then we will need to include those here.

modules/installation/proc-install-rhdh-airgapped-environment-ocp-operator.adoc
rm3l: @linfraze Has this been added? I might have missed it.

modules/installation/proc-install-rhdh-airgapped-environment-ocp-operator.adoc
linfraze: > if the mirror registry requires authentication.

so there are only certain conditions under which authentication is required? what are those conditions?

modules/installation/proc-install-rhdh-airgapped-environment-ocp-operator.adoc
linfraze: @rm3l had to switch tasks for a while but this update has been added now

modules/installation/proc-install-rhdh-airgapped-environment-ocp-operator.adoc
```diff
@@ -5,68 +5,146 @@
...
+
+include::../../artifacts/snip-rhdh-install-operator-next-steps.adoc[]
```
linfraze, line 73: @docs - is there an issue with using the same snippet more than once within a single file? it's not showing up in both places and the ccutil build is failing.

modules/installation/proc-install-rhdh-airgapped-environment-ocp-operator.adoc
rm3l: @linfraze This depends on how the environment is set up. But I assume most mirror registries would require some authentication (at least for registries reachable from the internet, like in partially disconnected environments).

rm3l (commented):
@linfraze LGTM overall. I wanted to take a quick look at the PR preview website, but it looks like the Workflow is failing: https://github.com/redhat-developer/red-hat-developers-documentation-rhdh/actions/runs/13725717386/job/38391563412?pr=935
Could you PTAL? Thanks.

modules/installation/proc-install-rhdh-airgapped-environment-ocp-operator.adoc
linfraze: Created a Jira issue to add a conditional prereq to deployment docs - https://issues.redhat.com/browse/RHIDP-6367

modules/installation/proc-install-rhdh-airgapped-environment-ocp-operator.adoc
linfraze: removing one of the snippets fixed the build error

rm3l (approved):
LGTM after reviewing the technical content.

/lgtm

modules/installation/proc-install-rhdh-airgapped-environment-ocp-operator.adoc
```diff
@@ -5,68 +5,152 @@
...
-* The jq package is installed. See link:https://jqlang.github.io/jq/download/[Download jq].
-* Podman is installed. See link:https://podman.io/docs/installation[Podman Installation Instructions].
-* Skopeo version 1.14 or higher is installed. link:https://github.com/containers/skopeo/blob/main/install.md[See Installing Skopeo].
-* If you already have a mirror registry for your cluster, an active Skopeo session with administrative access to this registry is required. See link:https://github.com/containers/skopeo#authenticating-to-a-registry[Authenticating to a registry] and link:https://docs.redhat.com/en/documentation/openshift_container_platform/{ocp-version}/html-single/disconnected_installation_mirroring/index#prerequisites_installing-mirroring-installation-images[Mirroring images for a disconnected installation].
+* You have installed Podman 5.3 or later. For more information, see link:https://podman.io/docs/installation[Podman Installation Instructions].
+* You have installed Skopeo 1.17 or later.
+* You have installed yq 4.44 or later.
```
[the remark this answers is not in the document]
linfraze: @reestr running the bash script (in place of performing the manual steps required to install the Operator in an air-gapped environment) requires the user to install several tools that (afaik) are not owned or maintained by Red Hat, including `skopeo`, `yq`, `umoci`. Providing any information about these tools or how to install them requires us to link to third-party websites (GitHub repos) that are also not owned or maintained by Red Hat.

What are your thoughts on using bash scripts to perform RHDH installation procedures? Does this approach cross the boundaries of RH support?

modules/installation/proc-install-rhdh-airgapped-environment-ocp-operator.adoc
```diff
@@ -5,68 +5,152 @@
...
+** link:https://docs.redhat.com/en/documentation/red_hat_developer_hub/1.4/html/installing_red_hat_developer_hub_on_microsoft_azure_kubernetes_service/proc-rhdh-deploy-aks-operator_title-install-rhdh-aks[Installing {product-short} on {aks-short} with the Operator]
+** link:https://docs.redhat.com/en/documentation/red_hat_developer_hub/1.4/html/installing_red_hat_developer_hub_on_openshift_dedicated_on_google_cloud_platform/proc-install-rhdh-osd-gcp-operator_title-install-rhdh-osd-gcp[Installing {product-short} on {gcp-short} with the Operator]
+** link:https://docs.redhat.com/en/documentation/red_hat_developer_hub/1.4/html/installing_red_hat_developer_hub_on_google_kubernetes_engine/proc-rhdh-deploy-gke-operator.adoc_title-install-rhdh-gke#proc-deploy-rhdh-instance-gke.adoc_title-install-rhdh-gke[Deploying {product-short} on {gke-short} with the Operator]
+
+== Installing {product} in a fully disconnected environment with the Operator
```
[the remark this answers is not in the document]
linfraze: +1 to more "why" info in general, and we can definitely provide some more conceptual information here, but using a fully disconnected environment is often a requirement rather than a choice. Some customers, like government agencies, must use a fully disconnected environment due to their high-security needs and are likely already knowledgeable about their technical environment and why it is needed.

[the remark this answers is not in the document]
hmanwani-rh: We’ve previously received feedback that our documentation often assumes users are already knowledgeable, and we simply present the information. Since security is a key reason users might choose a disconnected environment, how about we include something like:
"_In environments without internet access — whether for security, compliance, or operational reasons — a fully disconnected installation ensures that {product} can run reliably without external dependencies._"

This could help provide context for less experienced users, giving them a clearer understanding of why they might need this setup or why it’s beneficial.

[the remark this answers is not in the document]
linfraze: Sure, I'll paste it in

modules/installation/artifacts
```diff
@@ -0,0 +1 @@
+../../artifacts
```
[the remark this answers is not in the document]
linfraze: I've removed the symlink, but as a result, I have also had to remove the snippet that facilitates content reuse. In the future, let's address why snippets are not working without the use of the symlink and figure out what we can do differently.

jmagak (approved):

PR #1079, opened 2025-04-16
RHIDP-6570: Document the permission support to RBAC plugin

hmanwani-rh:

**Version(s):**
1.6
**Issue:**
[RHIDP-6570](https://issues.redhat.com/browse/RHIDP-6570)
**Preview:**

### DIFF
--- a/modules/authorization/proc-delegating-rbac-access.adoc
+++ b/modules/authorization/proc-delegating-rbac-access.adoc
@@ -1,9 +1,11 @@
 [id='proc-delegating-rbac-access_{context}']
 = Delegating role-based access controls (RBAC) access in {product}
 
-An enterprise customer requires the ability to delegate role-based access control (RBAC) responsibilities to individual team leads. In this scenario, you, as the administrator, can provide access to the RBAC plugin specifically to designated users, such as team leads. Each team lead is then able to manage permissions exclusively for users within their respective team or department, without visibility into or control over permissions outside their assigned scope. 
+An enterprise customer requires the ability to delegate role-based access control (RBAC) responsibilities to other individual in the organization. In this scenario, you, as the administrator, can provide access to the RBAC plugin specifically to designated users, such as team leads. Each team lead is then able to manage permissions exclusively for users within their respective team or department, without visibility into or control over permissions outside their assigned scope. This approach allows team leads to manage access and permissions for their own teams independently, while administrators maintain global oversight.
 
-The expected results of delegating RBAC access are as follows:
+In {product-very-short}, you can delegate RBAC access using the multitenancy feature of RBAC plugin, specifically the `IS_OWNER` conditional rule.
+
+By delegating the RBAC access, you can expect the following outcomes:
 
 * Team leads can manage RBAC settings for their teams independently.
 * Visibility of other users' or teams' permissions is restricted.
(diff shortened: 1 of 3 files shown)


PatAKnight (commented):
Looks good so far, but we probably should include more information about what conditional rule to use and who to set it to. As well as what specific steps the team lead will need to take to ensure that they can see the roles and permissions that they create. Here is some [docs](https://github.com/backstage/community-plugins/blob/main/workspaces/rbac/plugins/rbac-backend/docs/multitenancy.md) that I have written for the RBAC backend plugin upstream.

modules/authorization/proc-delegating-rbac-access.adoc
```diff
@@ -0,0 +1,179 @@
...
+.Example of granting the team lead role permission to create RBAC policies and read catalog entities
+[source,bash]
+----
+curl -X POST 'http://localhost:7007/api/permission/policies' \
+--header "Authorization: Bearer $ADMIN_TOKEN" \
+--header "Content-Type: application/json" \
+--data '[
+  {
+    "entityReference": "role:default/team_lead",
+    "permission": "policy-entity",
```
PatAKnight, line 54:
```suggestion
    "permission": "policy.entity.create",
```

PatAKnight (commented):
Fairly minor, and I need to fix it in the upstream docs as well, but I am working on removing the resource type from the `policy.entity.create` permission. This means that at some point in the future, admins will no longer be able to create permissions using `policy-entity, create` and will instead need to use `policy.entity.create, create`. 

I was hoping that we could use your PR to make that change. It looks like the create policy permission is only referenced in [one other location](https://github.com/redhat-developer/red-hat-developers-documentation-rhdh/blob/main/modules/authorization/ref-rbac-permission-policies.adoc?plain=1#L161). Do you mind updating it as well to remove the `policy-entity` from `policy.entity.create`?

PatAKnight (commented):
Small request

modules/authorization/proc-delegating-rbac-access.adoc
```diff
@@ -0,0 +1,179 @@
+[id='proc-delegating-rbac-access_{context}']
+= Delegating role-based access controls (RBAC) access in {product}
+
+An enterprise customer requires the ability to delegate role-based access control (RBAC) responsibilities to other individual in the organization. In this scenario, you, as the administrator, can provide access to the RBAC plugin specifically to designated users, such as team leads. Each team lead is then able to manage permissions exclusively for users within their respective team or department, without visibility into or control over permissions outside their assigned scope. This approach allows team leads to manage access and permissions for their own teams independently, while administrators maintain global oversight.
```
jmagak, line 4:
```suggestion
An enterprise customer requires the ability to delegate role-based access control (RBAC) responsibilities to other individuals in the organization. In this scenario, you, as the administrator, can provide access to the RBAC plugin specifically to designated users, such as team leads. Each team lead is then able to manage permissions exclusively for users within their respective team or department, without visibility into or control over permissions outside their assigned scope. This approach allows team leads to manage access and permissions for their own teams independently, while administrators maintain global oversight.
```

modules/authorization/proc-delegating-rbac-access.adoc
```diff
@@ -0,0 +1,179 @@
...
+
+.Verification
+* Log in as a team lead and verify the following:
++
+--
+** The RBAC UI is accessible.
+** Only the assigned users or group is visible.
+** Permissions outside the scoped team are not viewable or editable.
+--
+* Log in as an administrator and verify that you retain full visibility and conrol
```
jmagak, line 176:
```suggestion
* Log in as an administrator and verify that you retain full visibility and control
```

modules/authorization/proc-delegating-rbac-access.adoc
```diff
@@ -0,0 +1,179 @@
...
+  "metadata": {
+    "description": "This is an example team_a role",
+    "owner": "user:default/team_lead"
+  }
+}'
+----
++
+[NOTE]
+====
+You can set the ownership during creation but you can also update the ownership at any time.
```
jmagak, line 134:
```suggestion
You can set the ownership during creation, but you can also update the ownership at any time.
```

modules/authorization/proc-delegating-rbac-access.adoc
hmanwani-rh: Added a period (.) too

PatAKnight (approved):
Looks pretty good to me

PR #1101, opened 2025-04-25
RHIDP-7038: Doc air-gapped installation on K8s platforms with Helm

linfraze:

**Version(s):**
1.6
**Issue:**
https://issues.redhat.com/browse/RHIDP-7038
**Preview:**
https://redhat-developer.github.io/red-hat-developers-documentation-rhdh/pr-1101/install-rhdh-air-gapped/


modules/installation/proc-install-rhdh-airgapped-full-k8s-helm.adoc
```diff
@@ -0,0 +1,234 @@
...
+helm repo add _<helm_chart_repo_name>_ https://charts.openshift.io/
+helm repo update
+helm show values _<helm_chart_repo_name>_/redhat-developer-hub --version _<rhdh_version>_  values.default.yaml
+helm pull _<helm_chart_repo_name>_/redhat-developer-hub --version _<rhdh_version>_
+----
++
+where
+
+_<helm_chart_repo_name>_ :: Specifies the name of the Helm chart repository, for example, `openshift-helm-charts`.
+_<rhdh_version>_ :: Specifies the {product} version that you want to use, for example, `1.4.0`.
```
Fortune-Ndlovu, line 29: Do we have a product version variable instead hard-coding `1.4.0`

modules/installation/proc-install-rhdh-airgapped-full-k8s-helm.adoc
```diff
@@ -0,0 +1,234 @@
...
+skopeo login registry.redhat.io
+skopeo copy --all docker://${RHDH_IMAGE}:${RHDH_DIGEST} dir:./rhdh-hub
+skopeo copy --all docker://${PG_IMAGE}:${PG_DIGEST} dir:./postgresql
+----
+
+. Transfer the following files and directories to your air-gapped environment:
++
+* rhdh-hub
+* postgresql
+* Helm chart archive file, for example, `redhat-developer-hub-1.4.0.tgz`
```
Fortune-Ndlovu, line 58: We could add the product version variable here also, and for the remaining versions, if we have a product version.

modules/installation/proc-install-rhdh-airgapped-full-k8s-helm.adoc
```diff
@@ -0,0 +1,234 @@
...
++
+. Load the images onto the air-gapped host by running the following command:
```
Fortune-Ndlovu, line 60:
```suggestion
. Load the images onto the air-gapped host by running the following commands:
```

modules/installation/proc-install-rhdh-airgapped-partial-k8s-helm.adoc
```diff
@@ -0,0 +1,210 @@
...
+  volumePermissions:
+    enabled: true
+----
++
+. Install the Helm chart in the current namespace by running the following command:
++
+[source,terminal,subs="attributes+"]
+----
+helm install rhdh ./redhat-developer-hub-1.4.0.tgz -f values.yaml
+----
```
Fortune-Ndlovu, line 210: Do we also want to add the following:
```bash
+
where

_<helm_chart_archive_file_name>_ :: Specifies the name of the Helm chart archive file, for example, `redhat-developer-hub-1.4.0.tgz`.
```

Of course, then we would need to initially add the following instead:
```bash
----
helm install rhdh ./_<helm_chart_archive_file_name>_ -f values.yaml
----
```

titles/install-rhdh-air-gapped/master.adoc
```diff
@@ -1,14 +1,18 @@
 [id="title-install-rhdh-air-grapped"]
 include::artifacts/attributes.adoc[]
 :title: Installing {product} in an air-gapped environment
-:subtitle: Running {product} on {ocp-brand-name} in a networik restricted environment by using either the Operator or Helm chart.
-:abstract: Platform administrators can configure roles, permissions, and other settings to enable other authorized users to deploy an air-gapped {product-short} instance on any supported platform using either the Operator or Helm chart.
+:subtitle: Running {product} on {ocp-brand-name} in a networik restricted environment by using either the Operator or Helm chart
```
Fortune-Ndlovu, line 4: Do we want to leave the periods?

modules/installation/proc-install-rhdh-airgapped-full-k8s-helm.adoc
```diff
@@ -0,0 +1,234 @@
...
++
+[source,terminal,subs="+quotes"]
+----
+skopeo copy --all dir:./rhdh-hub docker://_<mirror_registry_name>_/_<rhdh_repo_name>_:${RHDH_DIGEST}
+
+skopeo copy --all dir:./postgresql docker://_<mirror_registry_name>_/_<postgresql_repo_name>_:${PG_DIGEST}
```
Fortune-Ndlovu, line 66: Do we want to always use  `<angle brackets>` for user-provided variables like `<mirror_registry_name>`, not underscores

modules/installation/proc-install-rhdh-airgapped-partial-k8s-helm.adoc
```diff
@@ -0,0 +1,210 @@
...
+.Procedure
+
+. In a terminal, download and extract the Helm chart by running the following commands:
++
+[source,terminal,subs="attributes+"]
+----
+helm repo add _<helm_chart_repo_name>_ https://charts.openshift.io/
+helm repo update
+helm pull _<helm_chart_repo_name>_/redhat-developer-hub --version _<rhdh_version>_
+helm show values _<helm_chart_repo_name>_/redhat-developer-hub --version _<rhdh_version>_  values.default.yaml
```
Fortune-Ndlovu, line 24:
```suggestion
helm show values <helm_chart_repo_name>/redhat-developer-hub --version <rhdh_version> > values.default.yaml
```

missing `>` redirect

modules/installation/proc-install-rhdh-airgapped-full-k8s-helm.adoc
```diff
@@ -0,0 +1,234 @@
...
+
+.Procedure
+
+. In a terminal, fetch the Helm charts values by running the following commands:
++
+[source,terminal,subs="+quotes"]
+----
+helm repo add _<helm_chart_repo_name>_ https://charts.openshift.io/
+helm repo update
+helm show values _<helm_chart_repo_name>_/redhat-developer-hub --version _<rhdh_version>_  values.default.yaml
```
Fortune-Ndlovu, line 22:
```suggestion
helm show values <helm_chart_repo_name>/redhat-developer-hub --version <rhdh_version> > values.default.yaml
```
missing `>` redirect

modules/installation/proc-install-rhdh-airgapped-full-k8s-helm.adoc
```diff
@@ -0,0 +1,234 @@
...
++
+where
+
+_<mirror_registry_name>_ :: Specifies the name of the target mirror registry that you want to push the images to, for example, `registry.example.com`.
+
+_<rhdh_repo_name>_ :: Specifies the name of the repository where your {product} image is stored, for example, `rhdh/rhdh-hub-rhel9`. This value must match the name of the {product} image that you loaded onto the air-gapped host.
+
+_<postgresql_repo_name>_ :: Specifies the name of the repository where your PostgreSQL image is stored, for example, `rhdh/postgresql-15`.
++
+* For {aks-short}, use the following `values.yaml` file template:
```
Fortune-Ndlovu, line 104: > About Dropping in values.yaml templates

IMO, yes, including platform-specific values.yaml templates (AKS, EKS, GKE) in the doc is a good idea because it makes it immediately consumable, not to mention, saves users time instead of guessing/creating their own, and it also makes it clear what settings (ingress, NodePort, fsGroup, etc.) matter per platform. It may seem long to our eyes, but it's super helpful for users reading this for the first time.

modules/installation/proc-install-rhdh-airgapped-full-k8s-helm.adoc
```diff
@@ -0,0 +1,234 @@
+[id="proc-install-rhdh-airgapped-full-k8s-helm_{context}"]
+= Installing {product} on a supported Kubernetes platform in a fully disconnected environment with the Helm chart
+
+In environments without internet access, a fully disconnected installation ensures that {product} can run reliably without external dependencies. This approach involves mirroring images and transferring them manually to the air-gapped environment.
+
+.Prerequisites
+
+* Skopeo 1.17 or later
+* Yq 4.4 or later
+* Access to registry.redhat.io for pulling images by using Skopeo login
```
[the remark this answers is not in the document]
linfraze: @Fortune-Ndlovu thanks! It is most important to be consistent with other RHDH docs, but if we want to make all RHDH docs consistent with `kubectl` then we can globally update these references across all RHDH docs later.

modules/installation/proc-install-rhdh-airgapped-full-k8s-helm.adoc
linfraze: We do, but this is just an example. Good idea to insert the attribute here, but it isn't entirely necessary.

titles/install-rhdh-air-gapped/master.adoc
linfraze: No, we do not use punctuation in a subtitle

modules/installation/proc-install-rhdh-airgapped-partial-k8s-helm.adoc
linfraze: The `_` italicizes the variable, which we do to be compliant with CCS guidelines

modules/installation/proc-install-rhdh-airgapped-full-k8s-helm.adoc
linfraze: Same comment / Q

modules/installation/proc-install-rhdh-airgapped-full-k8s-helm.adoc
```diff
@@ -0,0 +1,234 @@
...
+.Procedure
+
+. In a terminal, fetch the Helm charts values by running the following commands:
++
+[source,terminal,subs="+quotes"]
+----
+helm repo add _<helm_chart_repo_name>_ https://charts.openshift.io/
+helm repo update
+helm show values _<helm_chart_repo_name>_/redhat-developer-hub --version _<rhdh_version>_  values.default.yaml
+helm pull _<helm_chart_repo_name>_/redhat-developer-hub --version _<rhdh_version>_
```
linfraze, line 23: @Fortune-Ndlovu I should have thought to ask sooner, but what does `redhat-developer-hub` represent in these commands? Wonder if we should replace with a variable and define what this part of the command is.

modules/installation/proc-install-rhdh-airgapped-partial-k8s-helm.adoc
```diff
@@ -0,0 +1,210 @@
...
+----
++
+where
+
+_<helm_chart_repo_name>_ :: Specifies the name of the Helm chart repository, for example, `openshift-helm-charts`.
+_<rhdh_version>_ :: Specifies the {product} version that you want to use, for example, `1.4.0`.
```
linfraze, line 30: @Fortune-Ndlovu we are saying product version here, but is it actually more technically accurate to say that this is the product chart version?

modules/installation/proc-install-rhdh-airgapped-full-k8s-helm.adoc
Fortune-Ndlovu: Okay, thanks for the context!

modules/installation/proc-install-rhdh-airgapped-partial-k8s-helm.adoc
Fortune-Ndlovu: sounds good, in this instance we should have:
`helm show values _<helm_chart_repo_name>_/redhat-developer-hub --version _<rhdh_version>_  > values.default.yaml`

..because we need the `>` to redirect the entire default `values.yaml` contents of the Helm Chart into the users own local file `values.default.yaml` so they can customize it.

modules/installation/proc-install-rhdh-airgapped-full-k8s-helm.adoc
Fortune-Ndlovu: `redhat-developer-hub` is our rhdh chart name inside the Helm chart repository. It tells Helm exactly which chart to pull or show values for. Remember, we are installing rhdh in an air-gap using the helm chart, which has rhdh within and uses it within its installation. You could create a variable...the name redhat-developer-hub will never change as its our chart name inside the Helm chart repository - https://github.com/openshift-helm-charts/charts/tree/main/charts/redhat/redhat/redhat-developer-hub

modules/installation/proc-install-rhdh-airgapped-partial-k8s-helm.adoc
Fortune-Ndlovu: @linfraze yes technically since we use `--version` right, to extract the Helm chart. we are referring to the Helm Chart version not directly the product (RHDH) version BUT we know that the chart version and product are aligned e.g., HDH 1.4.0 and the chart 1.4.0, but they are not guaranteed to always match [semver reasons](https://semver.org/).

IMO you can leave:
```bash
Specifies the {product} version that you want to use, for example, `1.4.0`.
```
as is, because in practice, the chart version and product version are aligned

modules/installation/proc-install-rhdh-airgapped-full-k8s-helm.adoc
linfraze: Ah, thank you! Since our Chart name is static rather than configurable, we will leave it as it is.

assemblies/assembly-install-rhdh-airgapped-environment-k8s-helm.adoc
```diff
@@ -0,0 +1,19 @@
+[id="assembly-install-rhdh-airgapped-environment-k8s-helm_{context}"]
```
themr0c, line 1: > Create assembly and module file names that accurately and closely reflect the title of the assembly or module.
> At the top of every module, provide an anchor in the format [id="filename_{context}"] where filename is the exact name of the file, without the file extension (.adoc) and a module prefix. 

See https://redhat-documentation.github.io/modular-docs/#module_anchor-and-file-names-concept

linfraze: The id does match the filename; what is the issue?

themr0c:
```suggestion
[id="installing-rhdh-on-a-supported-kubernetes-platform-in-an-air-gapped-environment-with-the-helm-chart_{context}"]
```

modules/installation/proc-install-rhdh-airgapped-full-k8s-helm.adoc
```diff
@@ -0,0 +1,234 @@
+[id="proc-install-rhdh-airgapped-full-k8s-helm_{context}"]
```
themr0c, line 1:
```suggestion
[id="installing-rhdh-on-a-supported-kubernetes-platform-in-a-fully-disconnected-environment-with-the-helm-chart_{context}"]
```

+ rename the file to `proc-installing-rhdh-on-a-supported-kubernetes-platform-in-a-fully-disconnected-environment-with-the-helm-chart.adoc`

modules/installation/proc-install-rhdh-airgapped-full-k8s-helm.adoc
```diff
@@ -0,0 +1,234 @@
...
+
+In environments without internet access, a fully disconnected installation ensures that {product} can run reliably without external dependencies. This approach involves mirroring images and transferring them manually to the air-gapped environment.
+
+.Prerequisites
+
+* Skopeo 1.17 or later
+* Yq 4.4 or later
+* Access to registry.redhat.io for pulling images by using Skopeo login
+* Access to the Kubernetes cluster with `kubectl` configured
+* Helm 3.13 or later installed on the air-gapped host
```
themr0c, line 12: Consider clarifiying more precisely in prerequisites that you require 2 hosts, and clarify which prerequisites are for which host.

Given the steps that I read in the procedure, it seems all prerequistes are for the mirroring host

* One mirroring host with Internet access, acting as a transistory storage for the external dependencies.
** The `skopeo` CLI tool + Access to registry.redhat.io
** The `kubectl` CLI tool + Access to the air-gapped Kubernetes cluster
** The `helm` CLI tool
** The `yq` CLI tool
* The air-gapped Kubernetes cluster.

modules/installation/proc-install-rhdh-airgapped-full-k8s-helm.adoc
```diff
@@ -0,0 +1,234 @@
...
+
+.Procedure
+
+. In a terminal, fetch the Helm charts values by running the following commands:
```
themr0c, line 16: Clarify where you run the command (the command is expected to fail on the air-gapped host):


```suggestion
. On the mirroring host, in a terminal, fetch the Helm charts values by running the following commands:
```

assemblies/assembly-install-rhdh-airgapped-environment-k8s-helm.adoc
linfraze: I think this is fine as-is. The main thing is that the filename and id match, which they do.

modules/installation/proc-install-rhdh-airgapped-full-k8s-helm.adoc
linfraze: That is very long. The guideline also doesn't state that the filename and the title of the doc must match exactly. I think it is fine as-is and follows a naming scheme that works relative to the other files on this topic. The filename and id also match, which is the main requirement.

modules/installation/proc-install-rhdh-airgapped-full-k8s-helm.adoc
Fortune-Ndlovu: IMO, keeping consistency with how other air-grabbed docs are structured is a logical approach.

linfraze: @Fortune-Ndlovu can you confirm the technical accuracy of the change, please?

Gerry-Forde (approved):

PR #1218, opened 2025-06-03
RHIDP-7599: Update Template editor

pabel-rh:

**Version(s):**
main, 1.6.x
**Issue:**
RHIDP-7599, RHIDP-7436
**Preview:**
https://redhat-developer.github.io/red-hat-developers-documentation-rhdh/pr-1218/customizing/#proc-creating-templates_configuring-templates

### DIFF
--- a/modules/customizing-templates/proc-creating-templates.adoc
+++ b/modules/customizing-templates/proc-creating-templates.adoc
@@ -20,7 +20,7 @@ Alternately, you can use the Template Editor to do any of the following actions:
 
 To create a Software Template by using the Template Editor templates, complete the following steps:
 
-. In your {product} navigation menu, click *Catalog > Self-service*.
+. In your {product} navigation menu, click *Catalog > Self-service*. Alternatively, to go to *Self-service* page, in your header menu, click the (+) icon.
 . Click the overview menu and select *Manage Templates*.
 +
 image::rhdh/template-editor.png[Template Editor]
@@ -44,7 +44,7 @@ image::rhdh/template-editor.png[Template Editor]
 ... On the *Repository Location* tab, enter text into the required fields and click *Review*.
 ... Modify the YAML definition for the parameters of your template. For more information about these parameters, see <<Creating a Software Template as a YAML file>>.
 ... Review the information for accuracy, then click *Create*.
-.. When the Software Template is created, xref:proc-adding-templates_{context}[import your Software Template in your {product-very-short} instance].
+.. After the Software Template is created, xref:proc-adding-templates_{context}[import your Software Template in your {product-very-short} instance].
 
 .Verification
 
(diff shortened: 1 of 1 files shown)


jmagak:
Preview: https://redhat-developer.github.io/red-hat-developers-documentation-rhdh/pr-1218/

modules/customizing-templates/proc-creating-templates.adoc
```diff
@@ -37,7 +37,7 @@ image::rhdh/template-editor.png[Template Editor]
 * *Load Template Directory*: Use to load an existing `templates.yaml` file
 ** In your local file manager, navigate to the folder where your `templates.yaml` file is stored and click *Select*.
 * *Create New Template*: Use to create a `templates.yaml` file
-** In your local file manager, navigate to the folder where you want the Template Editor to create a `templates.yaml` file and click *Select*.
+.. In your local file manager, navigate to the folder where you want the Template Editor to create a `templates.yaml` file and click *Select*.
 .. On the *Template Editor* page, select the `templates.yaml` file.
 .. (Optional) You can preview and test your template specifications.
```
linfraze, line 42: It seems as if we are saying that lines 43-47 are optional steps that a user only does if they want to preview and test their template specification, but in my experience of this UX, it seems these are not optional, they are just the steps required to create a template. 

I'm not sure if the optional step on line 42 is even necessary to call out here. Can we remove it or move it to a Verification section? 

Can we also fix the indention of these steps so that they are not tucked under the optional step on line 42? 

I will try to make these changes directly

linfraze: On the 47 - In general, we use the term _After_ instead of _When_ or _Once_ in this context. But specifically, we should remove line 47 entirely since a) the importing step is already captured in another doc and b) the importing step has been added to a _Next steps_ section.

its-mitesh-kumar (approved):

modules/customizing-templates/proc-creating-templates.adoc
```diff
@@ -20,7 +20,7 @@ Alternately, you can use the Template Editor to do any of the following actions:
 
 To create a Software Template by using the Template Editor templates, complete the following steps:
 
-. In your {product} navigation menu, click *Catalog > Self-service*.
+. In your {product} navigation menu, click *Catalog > Self-service*. Alternatively, to go to *Self-service* page, in your header menu, click the (+) icon.
 . Click the overview menu and select *Manage Templates*.
```
invincibleJai, line 24: NIT: one suggestion,  not sure if the overview menu will be understood by all, i.e, what it is and where it is. 

@ShiranHi is there a term for the overview menu (Three dots on right) that is well-known, or is the overview menu fine

And maybe it'll make more intuitive for users to see screenshot of GIF

image

ShiranHi: I'm aware of **Overflow menu**, or **Kebab menu**

invincibleJai (approved):
/lgtm

PR #1259, opened 2025-07-06
RHIDP-5152: Add Frontend Plugn Wiring Section

jmagak:

**Version(s):** 1.6, 1.7,  main

**Issue:** [RHIDP-5152](https://issues.redhat.com/browse/RHIDP-5152)

**Preview:**


modules/dynamic-plugins/proc-extending-internal-icon-catalog.adoc
```diff
@@ -0,0 +1,27 @@
...
+[source,yaml]
+----
+# dynamic-plugins-config.yaml
+plugins:
+  - plugin: <plugin_path_or_url>
+    disabled: false
+    pluginConfig:
+      dynamicPlugins:
+        frontend:
+          <package_name>: # same as `scalprum.name` key in plugin's `package.json`
```
gashcrumb, line 19: This one comment in the YAML is actually a really important detail that we should highlight in the docs.  Also probably this example shouldn't include this comment.  However again folks often are confused where the "package_name" value comes from, it's whatever "scalprum.name" is set to in the package.json, or what the tooling currently outputs when building a frontend dynamic plugin.

assemblies/dynamic-plugins/assembly-front-end-plugin-wiring.adoc
```diff
@@ -0,0 +1,55 @@
...
+= Front-end plugin wiring
+:context: assembly-front-end-plugin-wiring
+
+You can configure front-end plugins to do the following:
+
+* Extend the internal library of available icons
+* Declare a new full page by defining a new route
+* Extend to an existing page using router bindings
+* Use mount points within the application
+* Provide additional utility APIs or replace existing ones
```
gashcrumb, line 11: I'm not sure is it worth adding mention of the SignInPage, scaffolder or techdocs extensions etc in this list?  This seems like a high-level overview of features, so maybe it doesn't need to be an exhaustive list I guess.

PR #1283, opened 2025-07-24
RHIDP-6574: Document resource required (Helm-based) for air-gapped environment

jmagak:

**Version(s):** 1.7 

**Issue:** [RHIDP-6574](https://issues.redhat.com/browse/RHIDP-6574)

**Preview:** 

[1. Installing Red Hat Developer Hub with Orchestrator using Helm](https://redhat-developer.github.io/red-hat-developers-documentation-rhdh/pr-1283/orchestrator/#install-rhdh-orchestrator-helm_orchestrator-helm)
[1.1. Resource requirements for air-gapped environments](https://redhat-developer.github.io/red-hat-developers-documentation-rhdh/pr-1283/orchestrator/#con-airgapped-requirements-orchestrator_install-rhdh-orchestrator-helm)

### DIFF
--- a/assemblies/assembly-install-rhdh-airgapped-environment-orchestrator-helm.adoc
+++ b/assemblies/assembly-install-rhdh-airgapped-environment-orchestrator-helm.adoc
@@ -10,7 +10,7 @@ endif::[]
 
 :context: install-rhdh-airgapped-environment-orchestrator-helm
 
-You can install {product} in a fully disconnected or partially disconnected environment using the {product} Helm chart.
+You can install {product} in a fully disconnected or partially disconnected environment by using the {product} Helm chart.
 
 include::modules/orchestrator/proc-install-rhdh-orchestrator-airgapped-partial.adoc[leveloffset=+1]
 


modules/installation/con-airgapped-requirements-orchestrator.adoc
```diff
@@ -0,0 +1,60 @@
...
+
+| Knative Serving and Eventing
+| 1–2 vCPUs, 1.5 GiB memory
+
+| PostgreSQL or External DB
+| 1 vCPU, 1 GiB memory
+
+| Persistent storage
+| 5 GiB (minimum), adjust based on workflow logs and artifacts
+|===
```
rm3l, line 31: Just curious, where are these recommended values coming from? I guess these should not be only for air-gapped environments, but for any env requiring workflow orchestration, no?

modules/installation/con-airgapped-requirements-orchestrator.adoc
```diff
@@ -0,0 +1,60 @@
...
+
+   Pre-download the Helm charts for `redhat-developer-hub`, `orchestrator-infra`, and any dependent charts (for example, `bitnami` and `backstage`). Host them in an internal chart repository or file server.
+
+. **CRD and operator installation**
+
+   Download the required CRDs and YAML manifests for SonataFlow and Knative operators and apply them manually.
+
+. **Configure network policies**
+
+   Validate inter-pod communication and ensure internal DNS resolution to the mirrored image registry.
```
rm3l, line 49: Shouldn't the concrete steps/commands to do this be documented? Like, which version of the plugins, Helm chart to mirror exactly? We should also consider the following scenarios: partially disconnected vs fully disconnected environments. See https://docs.redhat.com/en/documentation/red_hat_developer_hub/1.6/html-single/installing_red_hat_developer_hub_in_an_air-gapped_environment/index#assembly-install-rhdh-airgapped-environment-ocp-helm_title-install-rhdh-air-grapped

jmagak:
Closing this PR, as we currently don’t have any air-gap instructions for the orchestrator to document yet.

PR #1318, opened 2025-08-20
Orchestrator in RHDH Updates

jmagak:

**Version(s):** 1.7

**Issue:** [RHIDP-7556](https://issues.redhat.com/browse/RHIDP-7556)

**Preview:**


pabel-rh (commented):
When we mention "Operator", are we talking about the "RHDH Operator". If yes, do we need to mention that explicitly. I do understand that we have other types of operators too, for which, we use the lowercase 'o'.

modules/orchestrator/proc-enable-orchestrator-plugin.adoc
```diff
@@ -9,31 +9,27 @@ You can enable the Orchestrator plugin in {product-very-short} by configuring dy
...
-       - package: "@redhat/backstage-plugin-orchestrator-backend-dynamic@1.6.0"
-         disabled: false
-         dependencies:
-            - ref: sonataflow
-       - package: "@redhat/backstage-plugin-scaffolder-backend-module-orchestrator-dynamic@1.6.0"
-         disabled: false
-      - package: "@redhat/backstage-plugin-orchestrator-form-widgets@${ORCHESTRATOR_PLUGIN_VERSION}"
-         disabled: false
+- package: "@redhat/backstage-plugin-orchestrator@<plugin-version>"
+    disabled: false
```
gazarenkov, line 18: Just to enable with default settings package and disabled is enough, no needs in other fields (integrity, pluginConfig)

modules/orchestrator/proc-enable-orchestrator-plugin.adoc
```diff
@@ -48,15 +44,15 @@ data:
     includes:
       - dynamic-plugins.default.yaml
     plugins:
-      - package: "@redhat/backstage-plugin-orchestrator@1.6.0"
+      - package: "@redhat/backstage-plugin-orchestrator@<plugin-version>"
```
gazarenkov, line 47: <plugin-version> should probably be a variable (with a value like 1.6.0 or 1.7.0) as for complete example?

jmagak: > 'plugin-version' should probably be a variable (with a value like 1.6.0 or 1.7.0) as for complete example?

@gazarenkov, We choose to use a user-replaceable value because of the possibility of the version changing in the future. However, since this is mentioning a specific example for the user, I will introduce the version itself.

modules/orchestrator/proc-install-components-for-orchestrator-plugin.adoc
```diff
@@ -71,4 +66,9 @@ $ ./plugin-infra.sh
 ----
 
 .{product-very-short} Orchestrator Infra Helm Chart
-You can use {product-very-short} Orchestrator Infra Helm Chart to install components for the Orchestrator plugins in empty clusters.
\ No newline at end of file
+You can use {product-very-short} Orchestrator Infra Helm Chart to install components for the Orchestrator plugins in empty clusters.
+
+[IMPORTANT]
+====
+The Orchestrator Infra Helm chart is suitable for production use. By default, it does *not* auto-install or auto-upgrade the Operator: you must approve the install plans.
```
rm3l, line 73:
```suggestion
The Orchestrator Infra Helm chart is suitable for production use. By default, it does *not* auto-install or auto-upgrade the required Serverless Operators: you or a cluster administrator must approve the install plans.
```

Just to clarify that this is only about the Serverless Operators, not the RHDH Operator.

modules/orchestrator/proc-install-components-for-orchestrator-plugin.adoc
```diff
@@ -71,4 +66,9 @@ $ ./plugin-infra.sh
 ----
 
 .{product-very-short} Orchestrator Infra Helm Chart
```
rm3l, line 68: The complete name of this chart (as displayed in the OCP web console) is: "Orchestrator Infrastructure for Red Hat Developer Hub" Helm Chart
https://github.com/redhat-developer/rhdh-chart/blob/release-1.7/charts/orchestrator-infra/Chart.yaml#L2

modules/orchestrator/proc-enable-orchestrator-plugin.adoc
```diff
@@ -9,31 +9,13 @@ You can enable the Orchestrator plugin in {product-very-short} by configuring dy
...
-       - package: "@redhat/backstage-plugin-orchestrator-backend-dynamic@1.6.0"
-         disabled: false
-         dependencies:
-            - ref: sonataflow
-       - package: "@redhat/backstage-plugin-scaffolder-backend-module-orchestrator-dynamic@1.6.0"
-         disabled: false
-      - package: "@redhat/backstage-plugin-orchestrator-form-widgets@${ORCHESTRATOR_PLUGIN_VERSION}"
-         disabled: false
+- package: "@redhat/backstage-plugin-orchestrator@<plugin_version>"
+    disabled: false
```
rm3l, line 18:
```suggestion
  disabled: false
```

Fixing the indentation.

modules/orchestrator/proc-install-components-for-orchestrator-plugin.adoc
```diff
@@ -71,4 +66,9 @@ $ ./plugin-infra.sh
 ----
```
rm3l, line 66: One comment on the `curl` command above. Why don't you provide the exact command that people can just copy-paste , by just replacing `${PRODUCT_VERSION}` with `1.7` (I guess you have a variable for that, right?).

modules/orchestrator/proc-install-components-for-orchestrator-plugin.adoc
```diff
@@ -15,7 +15,7 @@ To run the Orchestrator plugin successfully on {ocp-short}, you must install com
 +
 [NOTE]
 ====
-The {product} Operator provisions these components automatically. Alternatively, you can install them manually by using another Operator.
+You must preinstall the components, or install them by using one of the supported methods described in the following section (manual installation, helper script, or Helm chart).
```
rm3l, line 18: > one of the supported methods described in the following section (manual installation, helper script, or Helm chart).

🤔 So here, we are saying the helper script is a supported method, but at the same time, saying "Do not use it in production." ?

gazarenkov: That is what upstream doc says:
"This script provides a quick way to install the OpenShift Serverless infrastructure for the Orchestrator plugin. It is safe to use in empty clusters but should be used with caution in production clusters."

gazarenkov: @rm3l not sure I got your point, sorry. You mean the phrasing is not good (which might be the case) or we should not mention it here because it has to be applied with caution for prod (but is incomparable simpler than other) or what? :)

modules/orchestrator/proc-install-components-for-orchestrator-plugin.adoc
jmagak: @rm3l, we use user-replaceable values instead because of the possibility of the version changing in the future just to avoid having to always update this whenever such happens. I have added the `specify your production version `note for users.

rm3l: image

@jmagak Is the goal to have `{product}` replaced with RHDH? And what are users expected to specify here? 1.7, 1.7.0 or 1.7.1, or ... ?

jmagak: > Okay. So I guess `{product}` should be replaced in the [preview](https://redhat-developer.github.io/red-hat-developers-documentation-rhdh/pr-1318/orchestrator/#proc-install-components-for-orchestrator-plugin_orchestrator-rhdh) then? image

Sure, You are right. That is the case now. Thanks, @rm3l.

gazarenkov (approved):

rm3l (approved):
/lgtm

PR #1349, opened 2025-09-04
RHDHBUGS-422: fixed duplicate titles in tables

deerskindoll:

**Version(s):**
 main, 1.7.1
**Issue:**
 https://issues.redhat.com/browse/RHDHBUGS-422
**Preview:**

### DIFF
--- a/modules/authorization/ref-rbac-permission-policies.adoc
+++ b/modules/authorization/ref-rbac-permission-policies.adoc
@@ -33,8 +33,6 @@ g, user:default/myuser, role:default/myrole
 
 {product-short} supports following permission policies:
 
-Catalog permissions::
-
 .Catalog permissions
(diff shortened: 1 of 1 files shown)


deerskindoll:
question for peer review: should I fix the use of allow?

question for tech review: are Tekton, ArgoCD and Quay permissions only for users and not for roles?

themr0c:
It is out of scope of the current task, therefore you can open followup Jiras in RHDHBUGS.

modules/authorization/ref-rbac-permission-policies.adoc
```diff
@@ -219,7 +219,7 @@ conditions:
       key: type
       value: kubernetes-cluster
 ----
-
+.OCM permissions
```
themr0c:
```suggestion
OCM permissions::

```

PR #2046, opened 2026-04-14
[RHIDP-7871]: Add authentication troubleshooting reference modules

themr0c:

**Version(s):** 1.6
**Issue:** https://issues.redhat.com/browse/RHIDP-7871
**Preview:** N/A

## Summary

- Add reference module documenting 3 common "Login failed" errors with causes and solutions
- Add reference module documenting LDAP malformed entity envelope error with cause and solution
- Include both modules in the existing troubleshoot authentication issues assembly


JessicaJHee (approved):
LGTM! We can continue adding to this as needed :)

PR #2053, opened 2026-04-15
RHDHBUGS-2956: Remove stale link and update reference in documentation

jmagak:

**Version(s):** release-1.9-post-cqa

**Issue:** [RHDHBUGS-2956](https://redhat.atlassian.net/browse/RHDHBUGS-2956)

**Preview:**


kadel (dismissed):
This won't fix the problem :-/ 
The new file doesn't contain configuration, so the step 3. where we say "Copy the package configuration from the [dynamic-plugins.default.yaml](https://github.com/redhat-developer/rhdh/blob/main/dynamic-plugins.default.yaml) file." no longer makes sense with new file as the file only list packages, but not its configuration. We will have to re-think this guide

kadel:
for 1.9 release we can still link to https://raw.githubusercontent.com/redhat-developer/rhdh/refs/heads/release-1.9/dynamic-plugins.default.yaml, but for next releases we will need re-work this guide

nickboldt (approved):
/lgtm

PR #2063, opened 2026-04-15
chore: only run CQA analysis for main and future branches, not the older ones where we didn't implement CQA scripts; also checkout build scripts from the PR's base branch, not main, so we get the correct tooling

nickboldt:
1. Only run CQA analysis for main and future branches, not the older ones where we didn't implement CQA scripts
2. Checkout trusted build scripts from the PR's base branch, not main, so we get the correct tooling 

https://redhat.atlassian.net/browse/RHIDP-12990

**Version(s):**

**Issue:**

**Preview:**

### DIFF
--- a/.github/workflows/content-quality-assessment.yml
+++ b/.github/workflows/content-quality-assessment.yml
@@ -17,9 +17,8 @@ on:
       - 'build/scripts/**'
     branches:
       - main
-      - rhdh-1.**
-      - 1.**.x
-      - release-1.**
+      - release-1.10
+      - release-2.**
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.number || github.event.pull_request.head.ref }}
--- a/.github/workflows/content-quality-assessment.yml
+++ b/.github/workflows/content-quality-assessment.yml
@@ -17,6 +17,7 @@ on:
       - 'build/scripts/**'
     branches:
       - main
+      - release-1.9-post-cqa
       - release-1.10
       - release-2.**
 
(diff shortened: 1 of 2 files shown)


.github/workflows/content-quality-assessment.yml
```diff
@@ -17,9 +17,8 @@ on:
       - 'build/scripts/**'
     branches:
       - main
-      - rhdh-1.**
-      - 1.**.x
-      - release-1.**
+      - release-1.10
```
Gerry-Forde, line 21: @nickboldt We currently have a `release-1.9-post-cqa` branch which should be CQA compliant. This will replace the existing `release-1.9` branch in the next couple of weeks (or before the next 1.9 z-stream release), so I'd suggest keeping the CQA check for this branch and then we'll update this again when we switch to the CQA-compliance branch.
```suggestion
      - release-1.9-post-cqa
      - release-1.10
```

.github/workflows/pr.yml
```diff
@@ -23,17 +23,16 @@ on:
   # will prevent the job from running until it's approved manually by human intervention.
   pull_request_target:
     types: [opened, synchronize, reopened, ready_for_review]
-    branches: 
+    branches:
     - main
-    - rhdh-1.**
-    - 1.**.x
```
themr0c, line 29: Not checking that it builds at all is a dangerous path.

themr0c: I'd suggest skipping the cqa checks but not skipping ccutil and lychee.

nickboldt: but  those are dead branches! 

Anything up to and including 1.7 is EOL. 

We haven't used either of those version schemes since rhdh-1.0 and 1.2.x, which are MORE than 2yrs old.

Gerry-Forde: @themr0c We are still running non-CQA checks for the `release-1.** `and `release-2.**` branches, so I think this is safe to merge.

PR #2098, opened 2026-04-23
[RHIDP-13092] : update generate-supported-plugins to run weekly on 1.8, 1.9, a…

Omar-AlJaljuli:
…nd main

**Version(s):**
Updates workflow relating to release-1.8, release-1.9, main
**Issue:**
Fixes: https://redhat.atlassian.net/browse/RHIDP-13092

Update GH action generate-supported-plugins to run weekly on the release-1.8, release1.9 and main branches to keep our tables up to date.


.github/workflows/generate-supported-plugins-pr.yml
```diff
@@ -7,39 +7,73 @@ on:
         description: "Branch to run the script from"
         required: true
         default: "main"
+  # Weekly: Monday 04:00 UTC (~ Sunday 11pm Eastern / 8pm Pacific). Scheduled workflows use the default branch YAML.
+  schedule:
+    - cron: "0 4 * * 1"
```
kim-tsao, line 12: should we run this a little more frequently?  Would a daily run make sense especially for main?

.github/workflows/generate-supported-plugins-pr.yml
```diff
@@ -7,39 +7,73 @@ on:
...
+        run: |
+          if [ "$EVENT_NAME" = "schedule" ] || [ "$TARGET_BRANCH" = "$INPUT_BRANCH" ]; then
+            echo "run=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "run=false" >> "$GITHUB_OUTPUT"
+          fi
+
       - name: Checkout repository
+        if: steps.gate.outputs.run == 'true'
         uses: actions/checkout@v4
```
kim-tsao:
```suggestion
        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
```

* Update to v6 ([v6.0.2](https://github.com/actions/checkout/releases/tag/v6.0.2) specifically) for better secrets handling.   See https://github.com/actions/checkout#whats-new
* Pin digests per [GH Security best practices](https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions)

.github/workflows/generate-supported-plugins-pr.yml
```diff
@@ -7,39 +7,73 @@ on:
...
         run: echo "TIMESTAMP=$(date +'%Y%m%d-%H%M%S' -u)" >> $GITHUB_ENV
 
       - name: Run the script on branch
-        run: bash modules/dynamic-plugins-reference/rhdh-supported-plugins.sh -b ${{ github.event.inputs.branch }}
+        if: steps.gate.outputs.run == 'true'
+        run: bash ${{ matrix.script_path }} -b ${{ matrix.target_branch }}
 
       - name: Create Pull Request
+        if: steps.gate.outputs.run == 'true'
         uses: peter-evans/create-pull-request@v6
```
kim-tsao: Update to the latest version v8? https://github.com/peter-evans/create-pull-request/releases/tag/v8.1.1 
```suggestion
        uses: peter-evans/create-pull-request@5f6978faf089d4d20b00c7766989d076bb2fc7f1 # v8
```

.github/workflows/generate-supported-plugins-pr.yml
Omar-AlJaljuli: I dont mind, but is that necessary? How often do plugins upgrade or change status?

Omar-AlJaljuli: Worst case scenario, before doc freeze (if thats a thing) we can manually run the job

kim-tsao (changes requested):
Can you please update the description in [CLAUDE.md](https://github.com/redhat-developer/red-hat-developers-documentation-rhdh/blob/main/.claude/CLAUDE.md?plain=1#L39) to say this will now run on schedule?

It would be good to have a manual trigger as fallback but it seems like I don't have permissions in this repo to do so

.github/workflows/generate-supported-plugins-pr.yml
nickboldt: this can be run manually by someone to trigger the job via workflow_dispatch -- see line 4

docs don't update that often (plugin updates are rare) and running it frequently will just create noise

but we should definitely run it after branching for 1.10 as there will be a slew of new tags created at that point for both the 1.10 and main branches. 

I would ensure this runs at least monthly, every time we bump a BS version and new tags appear in overlays repo.

kim-tsao (approved):
/lgtm

PR #2125, opened 2026-04-28
RHDHBUGS-2827: attributes update

deerskindoll:

**Version(s):**
 
main, release-1.9
**Issue:**

https://redhat.atlassian.net/browse/RHDHBUGS-2827
**Preview:**


deerskindoll:
@nickboldt I removed the minor version from our antora attributes file but I'm wondering if you'll need to update something on your end too, so that when you push release-related updates, the change I just made doesn't get overwritten

kadel (commented):
As I mentioned in https://redhat.atlassian.net/browse/RHDHBUGS-2827 I think that this will create problems for users. Users will want to know  exact version down to the patch version.

kadel (changes requested):

deerskindoll:
> As I mentioned in https://redhat.atlassian.net/browse/RHDHBUGS-2827 I think that this will create problems for users. Users will want to know exact version down to the patch version.

gotcha! I was a blinded by the must-have tag. I updated the Jira too.

deerskindoll:
cancelling

PR #2198, opened 2026-05-15
RHIDP-12595: Update Lightspeed procedure for 1.10 default enablement

jmagak:

**Version(s):** 1.10, main

**Issue:** [RHIDP-13620](https://redhat.atlassian.net/browse/RHIDP-13620)

**Preview:**


pabel-rh:
@jmagak, If I understand this correctly, the purpose of this module is to help users understand how to use Lightspeed. If that's the case, we already have a module for that which we can reuse. 
Can we point this to : https://docs.redhat.com/en/documentation/red_hat_developer_hub/1.9/html/interacting_with_red_hat_developer_lightspeed_for_red_hat_developer_hub/get-ai-assisted-help-for-your-development-tasks_interacting-with-developer-lightspeed-for-rhdh
(Of course, there are 1.10 related changes happening to this section as well, so it will be updated with the latest UI tags)

jmagak:
> If that's the case, we already have a module for that which we can reuse. 
I've done the following with this PR:

- Removed the duplicate module proc-access-the-ai-powered-virtual-assistant.adoc
- Removed the Lightspeed reference entirely from the migration assembly

Since Lightspeed is now installed and enabled by default in 1.10, I believe it doesn't need to be part of the front-end system migration documentation. Users who want to learn how to use Lightspeed can refer to the existing "[Get AI-assisted help for your development tasks](https://docs.redhat.com/en/documentation/red_hat_developer_hub/1.9/html/interacting_with_red_hat_developer_lightspeed_for_red_hat_developer_hub/get-ai-assisted-help-for-your-development-tasks_interacting-with-developer-lightspeed-for-rhdh#manage-chats_get-ai-assisted-help-for-your-development-tasks)" documentation.
cc @debsmita1, @pabel-rh

debsmita1 (commented):
LGTM !

pabel-rh (approved):

PR #2249, opened 2026-05-27
RHDHBUGS-2776: Clarify NODE_ENV environment variable in Extensions procedure

themr0c:
**Version(s):** 1.9, 1.10

**Issue:** https://redhat.atlassian.net/browse/RHDHBUGS-2776

**Preview:** https://redhat-developer.github.io/red-hat-developers-documentation-rhdh/pr-2249/extend_installing-and-viewing-plugins-in-rhdh/#manage-plugins-by-using-extensions_extensions-in-rhdh (6.5.2)

## Summary
- Add a reusable IMPORTANT snippet (`snip-node-env-development-extensions-note.adoc`) explaining why `NODE_ENV=development` is required for Extensions, what it controls (config validation leniency, logging/caching behavior), and the trade-off of running it in production
- Include the snippet at both NODE_ENV mention points in `proc-configure-rhdh-to-install-plugins-by-using-extensions.adoc` (Operator CR and Helm values sections)

## Test plan
- [ ] Verify the IMPORTANT admonition renders correctly in both Operator and Helm sections
- [ ] Verify the snippet does not break the nested list structure
- [ ] Confirm ccutil build passes for `extend_installing-and-viewing-plugins-in-rhdh`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/modules/shared/snip-node-env-development-extensions-note.adoc
+++ b/modules/shared/snip-node-env-development-extensions-note.adoc
@@ -2,10 +2,9 @@
 
 [IMPORTANT]
 ====
-The Extensions feature requires `NODE_ENV=development` to enable runtime plugin installation.
+By default, `NODE_ENV` is set to `production`.
+Extensions requires `NODE_ENV=development` to enable UI-driven plugin installation, where the plugin updates the dynamic plugins configuration files.
+Restarts are still required to install and load plugins.
 
-Setting `NODE_ENV` to `development` also relaxes configuration validation, which means that plugins with incomplete or missing configuration do not prevent the application from starting.
-This setting might also affect logging verbosity and caching behavior in Node.js dependencies.
-
-Understand this trade-off when running `NODE_ENV=development` in a production deployment.
+Using `NODE_ENV=development` in a production deployment is not supported for production plugin management.
 ====
--- a/modules/shared/snip-node-env-development-extensions-note.adoc
+++ b/modules/shared/snip-node-env-development-extensions-note.adoc
@@ -1,0 +1,11 @@
+:_mod-docs-content-type: SNIPPET
+
+[IMPORTANT]
+====
+The Extensions feature requires `NODE_ENV=development` to enable runtime plugin installation.
+
+Setting `NODE_ENV` to `development` also relaxes configuration validation, which means that plugins with incomplete or missing configuration do not prevent the application from starting.
+This setting might also affect logging verbosity and caching behavior in Node.js dependencies.
+
+Understand this trade-off when running `NODE_ENV=development` in a production deployment.
+====
--- a/modules/shared/proc-configure-rhdh-to-install-plugins-by-using-extensions.adoc
+++ b/modules/shared/proc-configure-rhdh-to-install-plugins-by-using-extensions.adoc
@@ -94,6 +94,8 @@ spec:
   database:
     enableLocalDb: true
 ----
++
+include::snip-node-env-development-extensions-note.adoc[]
 ... Update your `dynamic-plugins-rhdh` config map to include your extensions configuration file, as follows:
 +
 [source,yaml]
@@ -133,6 +135,8 @@ upstream:
       - name: NODE_ENV
         value: development
 ----
++
+include::snip-node-env-development-extensions-note.adoc[]
 ... Click *Upgrade*
 
 .Verification


modules/shared/snip-node-env-development-extensions-note.adoc
```diff
@@ -0,0 +1,11 @@
+:_mod-docs-content-type: SNIPPET
+
+[IMPORTANT]
+====
+The Extensions feature requires `NODE_ENV=development` to enable runtime plugin installation.
```
PatAKnight: Kind of true, I would probably phrase it more that the Extensions plugins require `NODE_ENV=development` for UI-driven installations where the plugin will update dynamic plugins configuration files, restarts are still required to install and load plugins.

modules/shared/snip-node-env-development-extensions-note.adoc
```diff
@@ -0,0 +1,11 @@
...
+
+Setting `NODE_ENV` to `development` also relaxes configuration validation, which means that plugins with incomplete or missing configuration do not prevent the application from starting.
```
PatAKnight: I don't believe that this part is true. I haven't heard of an ability to relax configuration validation globally within Backstage.

modules/shared/snip-node-env-development-extensions-note.adoc
```diff
@@ -0,0 +1,11 @@
...
+This setting might also affect logging verbosity and caching behavior in Node.js dependencies.
+
+Understand this trade-off when running `NODE_ENV=development` in a production deployment.
```
PatAKnight: I believe that it is mentioned elsewhere so might not be needed here as well. But if not, I would think a comment along the lines of `NODE_ENV=development` is unsupported for production plugin management.

PatAKnight (commented):
Just some in-line comments and then a couple more additional comments. 

First, I think we should mention somewhere that the default is `NODE_ENV=production`. 

Second, we seem to be included additional things that aren't specific to the Extensions plugins, i.e. logging, caching. I would either outright remove them and put them in a more generic `snip-node-env-developement` or just rename the file entirely and aim to make this more generic.

PatAKnight (approved):
LGTM

modules/shared/snip-node-env-development-extensions-note.adoc
```diff
@@ -0,0 +1,10 @@
+:_mod-docs-content-type: SNIPPET
+
+[IMPORTANT]
+====
+By default, `NODE_ENV` is set to `production`.
```
mattreid, line 5: @PatAKnight Do you know if this is explicitly set, or inferred in the absence of having NODE_ENV set to something else? To me, there's a difference between having a visible NODE_ENV = production in a config or not seeing it, and RHDH assuming behavior since no overriding value is present, but both could be considered defaults.

Is that distinction worth considering as part of this change? or am I overthinking it?

mattreid: When we got a question about Node_Env which led to this Jira, they wanted to know if they needed to set it to production to match their intended usage, and the only reference in the docs for Node_Env was in this guide, which is focused on how it affects the Extensions catalog. 

When they looked through our code, they found this: https://github.com/redhat-developer/rhdh-plugins/blob/main/workspaces/extensions/plugins/extensions-backend/src/router.ts#L299 which injects a default of Development.

mattreid:
In the same [guide](https://docs.redhat.com/en/documentation/red_hat_developer_hub/1.8/html/installing_and_viewing_plugins_in_red_hat_developer_hub/rhdh-extensions-plugins_assembly-install-third-party-plugins-rhdh#proc-extensions-enabling-plugins-installation_rhdh-extensions-plugins), it also mentions updating compose.yaml, which might not be needed anymore, but I'll have to defer to someone more technical. 

It may not have always been present, but [rhdh-local's default.env sets Node_Env to Development](https://github.com/redhat-developer/rhdh-local/blob/main/default.env#L35). Would it also need to be set in compose, or if set in the environment variables, is that sufficient and that step is now redundant?

Edit: I created https://redhat.atlassian.net/browse/RHDHBUGS-3305 for this.

modules/shared/snip-node-env-development-extensions-note.adoc
themr0c: ... But then the step in the procedure is not necessary. How do you plan top move forward? 
* Keep `development` as default => rewrite docs accordingly.
* Set `production` as default.

kim-tsao: I've been reviewing a similar [PR](https://github.com/redhat-developer/red-hat-developers-documentation-rhdh/pull/2218) and I see two supported scenarios:

1.  A running dev instance (NODE_ENV=development) of RHDH that can be shared among other developers
I think NODE_ENV needs to be set by the customers so they can decide whether to run on `production` or `development`.  If unset, it might default to 'undefined'.   Our e2e tests show how it's set for [helm](https://github.com/redhat-developer/rhdh/blob/105732db744f983854a5cfd1cc8086d314efbe8e/.ci/pipelines/value_files/diff-values_showcase-rbac_AKS.yaml#L58-L59) and [operator](https://github.com/redhat-developer/rhdh/blob/105732db744f983854a5cfd1cc8086d314efbe8e/.ci/pipelines/resources/rhdh-operator/rhdh-start.yaml#L38-L39) deployments. 

2.  RHDH Local which is single use (always set to NODE_ENV=development)

kim-tsao: Maybe we can combine this effort under the other [PR](https://github.com/redhat-developer/red-hat-developers-documentation-rhdh/pull/2218) which re-orgs the information for better flow.  

We can have an overview paragraph to clarify what dev environments we support (deployed and single-use) and how NODE_ENV is used to distinguish runtime behaviour for certain features

pabel-rh (approved):

modules/shared/snip-node-env-development-extensions-note.adoc
themr0c: OK, then logically I should close this PR, link https://redhat.atlassian.net/browse/RHDHBUGS-2776 to https://redhat.atlassian.net/browse/RHDHBUGS-2241, and reassign https://redhat.atlassian.net/browse/RHDHBUGS-2776 to @Gerry-Forde. I ask @Gerry-Forde.

PatAKnight: @mattreid, I went looking and it seems like you are right. We do not set this explicitly set this anywhere. I could have sworn that we did set `NODE_ENV=production` but I am not seeing it anywhere. Might have been that I was thinking the inverse was true since RHDH Local sets `NODE_ENV=development`

PR #2260, opened 2026-05-28
[RHDHBUGS-758]: Add Operator procedure for air-gapped dynamic plugin install

themr0c:
> [!IMPORTANT]
> **Do Not Merge** - PR for review only. Target: upstream `main`.

**Version(s):** 1.9, 1.10+

**Issue:** [RHDHBUGS-758](https://redhat.atlassian.net/browse/RHDHBUGS-758)

**Preview:**  https://redhat-developer.github.io/red-hat-developers-documentation-rhdh/pr-2260/extend_installing-and-viewing-plugins-in-rhdh/#install-dynamic-plugins-in-an-air-gapped-environment-by-using-the-operator_install-dynamic-plugins-in-rhdh

## Summary

The existing air-gapped plugin install docs only cover the Helm chart method. This PR adds an Operator-specific procedure.

The procedure covers:
1. Creating a Secret with `.npmrc` content pointing to an internal NPM registry
2. Mounting the Secret into the `install-dynamic-plugins` init container via `spec.application.extraFiles.secrets` at `/opt/app-root/src/.npmrc.dynamic-plugins`
3. The init container reads the `.npmrc` automatically via the pre-configured `NPM_CONFIG_USERCONFIG` env var

Technical details sourced from the [upstream Operator docs](https://github.com/redhat-developer/rhdh-operator/blob/main/docs/dynamic-plugins.md).

## Test plan

- [ ] CQA checks pass
- [ ] Procedure renders correctly in HTML preview
- [ ] Placed after the existing Helm-based air-gapped reference in the plugins title
- [ ] YAML examples are valid and use correct attribute substitution

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/assemblies/extend_installing-and-viewing-plugins-in-rhdh/assembly-install-dynamic-plugins-in-rhdh.adoc
+++ b/assemblies/extend_installing-and-viewing-plugins-in-rhdh/assembly-install-dynamic-plugins-in-rhdh.adoc
@@ -20,6 +20,8 @@ include::../modules/shared/ref-example-helm-chart-configurations-for-dynamic-plu
 
 include::../modules/shared/ref-installing-dynamic-plugins-in-an-air-gapped-environment.adoc[leveloffset=+1]
 
+include::../modules/shared/proc-install-dynamic-plugins-in-an-air-gapped-environment-by-using-the-operator.adoc[leveloffset=+1]
+
 include::assembly-mirror-rhdh-dynamic-plugins-in-disconnected-environments.adoc[leveloffset=+1]
 
 ifdef::parent-context[:context: {parent-context}]
--- a/assemblies/extend_installing-and-viewing-plugins-in-rhdh/assembly-install-dynamic-plugins-in-rhdh.adoc
+++ b/assemblies/extend_installing-and-viewing-plugins-in-rhdh/assembly-install-dynamic-plugins-in-rhdh.adoc
@@ -18,7 +18,7 @@ include::../modules/shared/con-installing-dynamic-plugins-using-the-helm-chart.a
 
 include::../modules/shared/ref-example-helm-chart-configurations-for-dynamic-plugin-installations.adoc[leveloffset=+1]
 
-include::../modules/shared/ref-installing-dynamic-plugins-in-an-air-gapped-environment.adoc[leveloffset=+1]
+include::../modules/shared/ref-install-dynamic-plugins-in-an-air-gapped-environment-by-using-the-helm-chart.adoc[leveloffset=+1]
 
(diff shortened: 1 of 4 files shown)


modules/shared/proc-install-dynamic-plugins-in-an-air-gapped-environment-by-using-the-operator.adoc
```diff
@@ -0,0 +1,72 @@
+:_mod-docs-content-type: PROCEDURE
+
+[id="install-dynamic-plugins-in-an-air-gapped-environment-by-using-the-operator_{context}"]
+= Install dynamic plugins in an air-gapped environment by using the Operator
+
+[role="_abstract"]
+In an air-gapped environment, you can install external dynamic plugins by configuring a custom NPM registry for the {product-short} Operator. You create a Kubernetes secret containing a `.npmrc` file and mount it into the `install-dynamic-plugins` init container.
```
rh-tokeefe:
```suggestion
To install external dynamic plugins in an air-gapped environment, configure a custom NPM registry by mounting a secret into the {product-short} Operator.
```

rh-tokeefe (approved):
LGTM, one suggestion for you to consider.

assemblies/extend_installing-and-viewing-plugins-in-rhdh/assembly-install-dynamic-plugins-in-rhdh.adoc
```diff
@@ -20,6 +20,8 @@ include::../modules/shared/ref-example-helm-chart-configurations-for-dynamic-plu
 
 include::../modules/shared/ref-installing-dynamic-plugins-in-an-air-gapped-environment.adoc[leveloffset=+1]
```
rm3l: Two points:
- To be consistent with the new file you added, this one should probably be renamed into `proc-install-dynamic-plugins-in-an-air-gapped-environment-by-using-the-helm-chart.adoc`, no?
Even the title there is confusing as it reads: `1.5. Installing dynamic plugins in an air-gapped environment` and then the content is about Helm.

Image

- The issue ([RHDHBUGS-758](https://redhat.atlassian.net/browse/RHDHBUGS-758)) this PR solves is quite old and was written at a time when we didn't have OCI plugins, IIRC, but only NPM plugins. So this (and the new procedure you added) are relevant to NPM plugins only. I might have missed it, but do we have similar procedures for installing OCI plugins in air-gapped envs? If not, maybe the titles of the sections should be updated to reflect that they are only about NPM packages? Or are you also planning to include similar procedures for OCI plugins in a follow-up issue/PR?

themr0c: Oooh we might need another issue. I need to sync with @Gerry-Forde on this question.

PR #2134, opened 2026-04-29
[RHIDP-12522]: Document OpenSSF Scorecard integration for security metrics

themr0c:
**IMPORTANT: Do Not Merge - To be merged by Docs Team Only**

**Version(s):** 1.10
**Issue:** https://redhat.atlassian.net/browse/RHIDP-12522
**Preview:** https://redhat-developer.github.io/red-hat-developers-documentation-rhdh/pr-<PR_NUMBER>/observability_evaluate-project-health-using-scorecards/

## Summary
- Add new procedure for integrating OpenSSF security metrics using Scorecards (enable module, annotate entities, configure exclude/include, threshold details)
- Update metrics reference table with 18 OpenSSF checks and a new Thresholds column
- Update assembly to include the new OpenSSF procedure alongside GitHub and Jira

## Files changed
- **NEW** `modules/observability_evaluate-project-health-using-scorecards/proc-integrate-openssf-health-metrics-using-scorecards.adoc`
- **UPDATED** `modules/observability_evaluate-project-health-using-scorecards/ref-supported-scorecard-metrics-providers.adoc`
- **UPDATED** `assemblies/observability_evaluate-project-health-using-scorecards/assembly-install-and-configure-scorecards-to-view-metrics.adoc`

## Notes
- OCI version tag uses placeholder `<tag>` — will update once rhdh PR redhat-developer/rhdh#4707 merges and the exact version is known
- OpenSSF module is currently being promoted to the RHDH build (overlay merged, packages PR open)
- Dependabot module is out of scope (not shipping in 1.10)

PR #2313, opened 2026-06-03
[RHDHBUGS-3308]: Fix scorecard metrics exclude/include and extract shared procedures

themr0c:
> [!IMPORTANT]
> **Do Not Merge** – review only. Merge when approved by docs team lead.

## Summary

Addresses unresolved review comments from PR #2134 (RHIDP-12522):

1. **Fix per-entity annotation name:** `scorecard.io/exclude_metrics` -> `scorecard.io/disabled-metrics` (per @alizard0, confirmed by [test overlay](https://github.com/janus-qe/overlay-scorecard-plugin-test/blob/main/openssf-scorecard-only.yaml))
2. **Extract exclude/include into shared procedures:** The exclude/include steps were OpenSSF-specific but apply to all providers (per @dzemanov). Split into two new procedures:
   - *Exclude specific scorecard metrics per entity* — per-entity `scorecard.io/disabled-metrics` annotation
   - *Exclude or include scorecard metrics globally* — global `excludeMetrics`/`includeMetrics` app-config
3. **Add reference module:** Available scorecard metric IDs per provider (Filecheck, GitHub, Jira, OpenSSF) with descriptions
4. **Remove invalid global config** from OpenSSF procedure (per @dzemanov)

## Version(s):
- 1.10.0

## Issue:
- https://redhat.atlassian.net/browse/RHDHBUGS-3308
- https://redhat.atlassian.net/browse/RHDHBUGS-3211

## Preview:
- N/A


alizard0 (approved):
lgtm

modules/observability_evaluate-project-health-using-scorecards/proc-disable-scorecard-metrics-globally.adoc
```diff
@@ -0,0 +1,35 @@
...
+* You have {scorecard-plugin-book-link}#set-up-scorecards-to-monitor-your-project-health_{context}[set up the Scorecard plugin].
+* You have {configuring-book-link}[added a custom {product-short} application configuration].
+
+.Procedure
+
+* Add the following configuration to your {product-very-short} `{my-app-config-file}` file:
++
+[source,yaml,subs="+quotes,+attributes"]
+----
+scorecard:
```
alizard0, line 21: This is not correct:
1. disableMetrics instead of exclude.
2. there is no includeMetrics 

These are the options available. 

```
scorecard:
  disabledMetrics:
    - openssf.packaging
  entityOverrides:
    disabledMetrics:
      enabled: false
      except:
        - openssf.maintained
```

alizard0 (changes requested):

modules/observability_evaluate-project-health-using-scorecards/proc-disable-scorecard-metrics-globally.adoc
dzemanov, line 21: @themr0c can you fix the configuration according to the https://redhat.atlassian.net/browse/RHDHBUGS-3211 description? `excludeMetrics` and `includeMetrics` configuration keys don't exist

dzemanov (changes requested):

modules/observability_evaluate-project-health-using-scorecards/proc-disable-scorecard-metrics-globally.adoc
dzemanov: I would suggest to use other words than exclude and include metrics, as it might imply those metrics are just filtered out. When metrics are disabled in `app-config` or by entity annotations, the calculations are not run at all, they are disabled (globally or for that entity). The `entityOverrides.disabledMetrics` configuration determines whether users can disable metrics using annotations.

dzemanov: @alizard0 this configuration also doesn't exist:
```
 scorecard:
  plugins:
    openssf:
      enabled: true
```

alizard0: @themr0c this might help: https://github.com/redhat-developer/rhdh-plugins/blob/main/workspaces/scorecard/plugins/scorecard-backend/docs/disabled-metrics-logic.md

dzemanov (changes requested):
@themr0c when reviewing, I noticed that scorecard filechecks are under different section then all other scorecard metrics. Would you mind moving them from [7. Configure file-level checks to verify repositories contain required compliance documentation](https://redhat-developer.github.io/red-hat-developers-documentation-rhdh/pr-2313/observability_evaluate-project-health-using-scorecards/#configure-file-level-checks-to-verify-repositories-contain-required-compliance-documentation_evaluate-project-health-using-scorecards) under 4.4? \
[4. Install and configure Scorecards to view metrics](https://redhat-developer.github.io/red-hat-developers-documentation-rhdh/pr-2313/observability_evaluate-project-health-using-scorecards/#install-and-configure-scorecards-to-view-metrics_evaluate-project-health-using-scorecards)
[4.1. Integrate GitHub health metrics using Scorecards](https://redhat-developer.github.io/red-hat-developers-documentation-rhdh/pr-2313/observability_evaluate-project-health-using-scorecards/#integrate-github-health-metrics-using-scorecards_install-and-configure-scorecards-to-view-metrics)
[4.2. Integrate Jira health metrics using Scorecards](https://redhat-developer.github.io/red-hat-developers-documentation-rhdh/pr-2313/observability_evaluate-project-health-using-scorecards/#integrate-jira-health-metrics-using-scorecards_install-and-configure-scorecards-to-view-metrics)
[4.3. Integrate OpenSSF security metrics by using Scorecards](https://redhat-developer.github.io/red-hat-developers-documentation-rhdh/pr-2313/observability_evaluate-project-health-using-scorecards/#integrate-openssf-security-metrics-by-using-scorecards_install-and-configure-scorecards-to-view-metrics)

modules/observability_evaluate-project-health-using-scorecards/ref-available-scorecard-metric-ids.adoc
```diff
@@ -0,0 +1,40 @@
+:_mod-docs-content-type: REFERENCE
+
+[id="available-scorecard-metric-ids_{context}"]
+= Available scorecard metric IDs
+
+[role="_abstract"]
+Use the following metric IDs when excluding or including scorecard metrics per entity or globally.
```
dzemanov:
```suggestion
Use the following metric IDs when disabling scorecard metrics per entity or globally.
```

dzemanov: It would be nice to also rename procedure files from `exclude-or-include` and `proc-exclude` so we can easier locate this section about disabling metrics

dzemanov (changes requested):
Thank you @themr0c for the changes, I have left only a couple of small suggestions, otherwise looks great!

dzemanov:
@alizard0 is there docs issue for dependabot metrics? I haven't seen them in the scorecard documentation.

themr0c:
@alizard0 Quick question: the plugin's `config.d.ts` defines the config key as `scorecard.entityAnnotations.disabledMetrics`, but your review comment used `entityOverrides`. Can you confirm which key is correct? I want to make sure the docs match the actual config schema.

Reference: https://github.com/redhat-developer/rhdh-plugins/blob/main/workspaces/scorecard/plugins/scorecard-backend/config.d.ts

themr0c:
@alizard0 I created this issue for dependabot metrics: https://redhat.atlassian.net/browse/RHDHBUGS-3326

alizard0:
> @alizard0 Quick question: the plugin's `config.d.ts` defines the config key as `scorecard.entityAnnotations.disabledMetrics`, but your review comment used `entityOverrides`. Can you confirm which key is correct? I want to make sure the docs match the actual config schema.
> 
> Reference: https://github.com/redhat-developer/rhdh-plugins/blob/main/workspaces/scorecard/plugins/scorecard-backend/config.d.ts

I missed that when I was refactoring the terms. Anyways, I think this doc will clarify things: https://github.com/redhat-developer/rhdh-plugins/blob/main/workspaces/scorecard/plugins/scorecard-backend/docs/disabled-metrics-logic.md

@dzemanov

dzemanov:
@themr0c both keys are used, for different purposes. You have them correctly explained under where clause for both `disabledMetrics`, `entityOverrides.disabledMetrics.enabled` and `entityOverrides.disabledMetrics.except`.

Edit: Pardon, I missed the key being used was entityOverrides, yeah, it should be `entityAnnotations`. @themr0c thank you for spotting! Fixed it also in Jira issue description. So the keys should be: `entityAnnotations.disabledMetrics.enabled` and `entityAnnotations.disabledMetrics.except`

dzemanov:
@themr0c Do we want to create a separate page for listing metrics when there is already page for that: https://redhat-developer.github.io/red-hat-developers-documentation-rhdh/release-1.10/observability_evaluate-project-health-using-scorecards/#supported-scorecard-metrics-providers_evaluate-project-health-using-scorecards
Since then whenever we add new metrics, we will need to remember to update it in both places. It is true that it might be unclear for user when looking for what metric ids are available, to check Supported Scorecard metrics providers section. Maybe it can be renamed to Available scorecard metric providers and metric IDs?

modules/observability_evaluate-project-health-using-scorecards/proc-disable-scorecard-metrics-globally.adoc
```diff
@@ -0,0 +1,42 @@
...
+  disabledMetrics:
+    - openssf.packaging
+  entityOverrides:
```
dzemanov:
```suggestion
  entityAnnotations:
```

modules/observability_evaluate-project-health-using-scorecards/proc-disable-scorecard-metrics-globally.adoc
```diff
@@ -0,0 +1,42 @@
...
+      enabled: true
+      except:
+        - openssf.maintained
+----
++
+where:
+
+`disabledMetrics`:: Enter a list of metric IDs to disable globally. Metrics in this list are always skipped and cannot be overridden by entity annotations. For available values, see xref:available-scorecard-metric-ids_install-and-configure-scorecards-to-view-metrics[Available scorecard metric IDs].
+
+`entityOverrides.disabledMetrics.enabled`:: Enter `true` to allow entity owners to disable metrics by using the `scorecard.io/disabled-metrics` annotation. Enter `false` to prevent entity owners from disabling metrics. Defaults to `true`.
```
dzemanov:
```suggestion
`entityAnnotations.disabledMetrics.enabled`:: Enter `true` to allow entity owners to disable metrics by using the `scorecard.io/disabled-metrics` annotation. Enter `false` to prevent entity owners from disabling metrics. Defaults to `true`.
```

modules/observability_evaluate-project-health-using-scorecards/proc-disable-scorecard-metrics-globally.adoc
```diff
@@ -0,0 +1,42 @@
...
+
+`entityOverrides.disabledMetrics.except`:: Enter a list of metric IDs that entity owners cannot disable, even when `enabled` is `true`. Use this to ensure critical metrics always run.
```
dzemanov:
```suggestion
`entityAnnotations.disabledMetrics.except`:: Enter a list of metric IDs that entity owners cannot disable, even when `enabled` is `true`. Use this to ensure critical metrics always run.
```

dzemanov (changes requested):

dzemanov:
> @alizard0 I created this issue for dependabot metrics: https://redhat.atlassian.net/browse/RHDHBUGS-3326

Thank you for creating the issue. I have updated a little its description.

dzemanov:
> @themr0c when reviewing, I noticed that scorecard filechecks are under different section then all other scorecard metrics. Would you mind moving them from [7. Configure file-level checks to verify repositories contain required compliance documentation](https://redhat-developer.github.io/red-hat-developers-documentation-rhdh/pr-2313/observability_evaluate-project-health-using-scorecards/#configure-file-level-checks-to-verify-repositories-contain-required-compliance-documentation_evaluate-project-health-using-scorecards) under 4.4? [4. Install and configure Scorecards to view metrics](https://redhat-developer.github.io/red-hat-developers-documentation-rhdh/pr-2313/observability_evaluate-project-health-using-scorecards/#install-and-configure-scorecards-to-view-metrics_evaluate-project-health-using-scorecards) [4.1. Integrate GitHub health metrics using Scorecards](https://redhat-developer.github.io/red-hat-developers-documentation-rhdh/pr-2313/observability_evaluate-project-health-using-scorecards/#integrate-github-health-metrics-using-scorecards_install-and-configure-scorecards-to-view-metrics) [4.2. Integrate Jira health metrics using Scorecards](https://redhat-developer.github.io/red-hat-developers-documentation-rhdh/pr-2313/observability_evaluate-project-health-using-scorecards/#integrate-jira-health-metrics-using-scorecards_install-and-configure-scorecards-to-view-metrics) [4.3. Integrate OpenSSF security metrics by using Scorecards](https://redhat-developer.github.io/red-hat-developers-documentation-rhdh/pr-2313/observability_evaluate-project-health-using-scorecards/#integrate-openssf-security-metrics-by-using-scorecards_install-and-configure-scorecards-to-view-metrics)

I have created a bug for this: https://redhat.atlassian.net/browse/RHDHBUGS-3319

jmagak (approved):
LGTM

PR #2609, opened 2026-09-16
RHDHBUGS-3781: RHDH 1.10 GitHub integration documentation uses an invalid value for `integrations.github.host`

jmagak:

**Version(s):** 1.10, 2.1

**Issue:** [RHDHBUGS-3781](https://redhat.atlassian.net/browse/RHDHBUGS-3781)

**Preview:**


modules/shared/proc-share-secrets-with-github.adoc
```diff
@@ -82,7 +82,12 @@ Enter the saved **Client Secret**.
 Enter the saved **Private key**.
 
 `GITHUB_URL`::
-Enter the GitHub host domain: `https://github.com`.
+Enter the GitHub instance URL with a scheme, such as `https://github.com`.
+The `githubOrg` provider `githubUrl` field expects a full URL.
```
karthikjeeyar: `GITHUB_URL` is a configuration is needed only for github-org plugin `@backstage/plugin-catalog-backend-module-github-org` plugin 

https://backstage.io/docs/integrations/github/org#installation.


I dont see github-org configuration in our documentation, so can we remove this to avoid confusion, wdyt?

jmagak: > I dont see github-org configuration in our documentation, so can we remove this to avoid confusion, wdyt?

Sure. I've removed the GITHUB_URL entry here. It's still kept where the githubOrg provider's githubUrl field is actually used (importing users and groups from GitHub).

karthikjeeyar (approved):
Thanks @jmagak. Looks good to me

pabel-rh (approved):

herbertkb:
In my opinion, there should be an addition to the `.catalog.provider.github.githubUrl` so that it can still use the `GITHUB_HOST` by pre-pending the scheme.

```
catalog:
  providers:
    github:
      githubUrl: https://${GITHUB_HOST}
```

